PR #2, opened 2022-09-11
Fix QR code bug when certain characters are in ssid or PSK

LoNicoM:
When < or > are in the psk it doesn't render these chars properly, also when ; in passkey QR code doesn't generate properly


pedro0311:
I'm sorry but it's only a mirror and I can't merge PRs here.
Could you please add it in the main repo: https://bitbucket.org/pedro311/freshtomato-arm/pull-requests/ ?
If you can't, I will add this as a separate commit.

Thanks.

pedro0311:
Already added on bitbucket

PR #3, opened 2023-12-30
stops SSL verifying for TransmissionBT

uDEV2019:
Sometimes a useful change...


uDEV2019:
sorry, that was wrong and not planned.

PR #4, opened 2024-06-17
Improve UPnP IGD & PCP/NAT-PMP

Self-Hosting-Group:
- Follow upstream change miniupnp/miniupnp@02da705
- Change default `upnp_ssdp_interval` to 900s to follow specification and upstream miniupnp/miniupnp@9339f0e

Commited in 31e715b

- Improve wording and UI, include `PCP` as supported and use `UPnP IGD` as described in more detail
- Remove `Show In My Network Places` option as icon is always shown in Windows
- Remove `Inactive Rules Cleaning` options as not standard/working
- Use secure mode setting for PCP
- Improve listing and rearrange table headers
- Invert/reword secure mode

Commited in 0ce92cb

- During testing I found a issue that creating port maps/forwards for IPv6 (using PCP) does not work, daemon complains that a required ip6tables chain does not exist
- Suggested to compile IGDv2 into the daemon but disable it at runtime, see comment in next PR for reasons

Fixed in bb80103

- Restart the daemon after IPv6 is available so that its support is not deactivated (log: `no HTTP IPv6 address, disabling IPv6`)

Fixed in b137790

freshtomato
_[Screenshot 2024.3 release](https://github.com/user-attachments/assets/57e42a93-38f2-49f1-afd1-b3418ea453c6)_

The **Port Control Protocol (PCP)** is the successor to NAT-PMP, has similar protocol concepts and packet formats, but adds support for **IPv6 port maps** and options/extensions. For more information, see:
**Port Mapping Protocols Overview and Comparison 2024: About UPnP IGD & PCP/NAT-PMP**
https://github.com/Self-Hosting-Group/wiki/wiki/Port-Mapping-Protocols-Overview

### DIFF
--- a/release/src-rt-6.x.4708/router/Makefile
+++ b/release/src-rt-6.x.4708/router/Makefile
@@ -1520,7 +1520,7 @@ miniupnpd/stamp-h1:
 	$(call patch_files,miniupnpd)
 	cd miniupnpd && \
 		PKG_CONFIG_LIBDIR="" \
-		./configure --leasefile --vendorcfg --portinuse --firewall=iptables --iptablespath=$(TOP)/$(IPTABLES_TARGET) $(if $(TCONFIG_IPV6),--ipv6,)
+		./configure --leasefile --vendorcfg --portinuse --igd2 --firewall=iptables --iptablespath=$(TOP)/$(IPTABLES_TARGET) $(if $(TCONFIG_IPV6),--ipv6,)
 	@touch $@
 
 miniupnpd: miniupnpd/stamp-h1
--- a/release/src-rt-6.x.4708/router/config/Config
+++ b/release/src-rt-6.x.4708/router/config/Config
@@ -162,7 +162,7 @@ config CONFIG_PPP
        default y
 
 config CONFIG_UPNP
-       bool "UPnP IGD server"
+       bool "UPnP IGD & PCP/NAT-PMP daemon"
        depends on CONFIG_NETCONF && CONFIG_NVRAM && CONFIG_SHARED && CONFIG_NAT
        default y
 
--- a/release/src-rt-6.x.4708/router/httpd/tomato.c
+++ b/release/src-rt-6.x.4708/router/httpd/tomato.c
@@ -954,7 +954,6 @@ static const nvset_t nvset_list[] = {
 	{ "upnp_secure",		V_01				},
 	{ "upnp_port",			V_RANGE(0, 65535)		},
 	{ "upnp_ssdp_interval",		V_RANGE(10, 9999)		},
-	{ "upnp_mnp",			V_01				},
 	{ "upnp_clean",			V_01				},
 	{ "upnp_clean_interval",	V_RANGE(60, 65535)		},
 	{ "upnp_clean_threshold",	V_RANGE(0, 9999)		},
--- a/release/src-rt-6.x.4708/router/rc/services.c
+++ b/release/src-rt-6.x.4708/router/rc/services.c
@@ -1823,7 +1823,7 @@ void start_upnp(void)
 	           "upnp_nat_postrouting_chain=pupnp\n"
 	           "notify_interval=%d\n"
 	           "system_uptime=yes\n"
-	           "friendly_name=FreshTomato UPnP daemon\n"
+	           "friendly_name=FreshTomato UPnP IGD &amp; PCP\n"
 	           "model_name=%s\n"
 	           "model_url=https://freshtomato.org/\n"
 	           "manufacturer_name=FreshTomato Firmware\n"
--- a/release/src-rt-6.x.4708/router/rc/services.c
+++ b/release/src-rt-6.x.4708/router/rc/services.c
@@ -1850,13 +1850,8 @@ void start_upnp(void)
 	else
 		fprintf(f, "clean_ruleset_interval=0\n");
 
-	if (nvram_get_int("upnp_mnp")) {
-		https = nvram_get_int("https_enable");
-		fprintf(f, "presentation_url=http%s://%s:%s/forward-upnp.asp\n", (https ? "s" : ""), nvram_safe_get("lan_ipaddr"), nvram_safe_get(https ? "https_lanport" : "http_lanport"));
-	}
-	else
-		/* Empty parameters are not included into XML service description */
-		fprintf(f, "presentation_url=\n");
+	https = nvram_get_int("https_enable");
+	fprintf(f, "presentation_url=http%s://%s:%s/forward-upnp.asp\n", (https ? "s" : ""), nvram_safe_get("lan_ipaddr"), nvram_safe_get(https ? "https_lanport" : "http_lanport"));
 
 	f_read_string("/proc/sys/kernel/random/uuid", uuid, sizeof(uuid));
 	fprintf(f, "uuid=%s\n", uuid);
--- a/release/src-rt-6.x.4708/router/rc/services.c
+++ b/release/src-rt-6.x.4708/router/rc/services.c
@@ -1817,6 +1817,7 @@ void start_upnp(void)
 	           "port=%d\n"
 	           "enable_upnp=%s\n"
 	           "enable_pcp_pmp=%s\n"
+	           "force_igd_desc_v1=yes\n"
 	           "secure_mode=%s\n"
 	           "pcp_allow_thirdparty=%s\n"
 	           "upnp_forward_chain=upnp\n"
--- a/release/src-rt-6.x.4708/router/rc/services.c
+++ b/release/src-rt-6.x.4708/router/rc/services.c
@@ -1818,6 +1818,7 @@ void start_upnp(void)
 	           "enable_upnp=%s\n"
 	           "enable_pcp_pmp=%s\n"
 	           "secure_mode=%s\n"
+	           "pcp_allow_thirdparty=%s\n"
 	           "upnp_forward_chain=upnp\n"
 	           "upnp_nat_chain=upnp\n"
 	           "upnp_nat_postrouting_chain=pupnp\n"
@@ -1834,6 +1835,7 @@ void start_upnp(void)
 	           (enable & 1) ? "yes" : "no",			/* upnp enable */
 	           (enable & 2) ? "yes" : "no",			/* natpmp enable */
 	           nvram_get_int("upnp_secure") ? "yes" : "no",	/* secure_mode (only forward to self) */
+	           nvram_get_int("upnp_secure") ? "no" : "yes",
 	           nvram_get_int("upnp_ssdp_interval"),
 	           nvram_safe_get("t_model_name"));
 
--- a/release/src-rt-6.x.4708/router/rc/services.c
+++ b/release/src-rt-6.x.4708/router/rc/services.c
@@ -1837,19 +1837,6 @@ void start_upnp(void)
 	           nvram_get_int("upnp_ssdp_interval"),
 	           nvram_safe_get("t_model_name"));
 
-	if (nvram_get_int("upnp_clean")) {
-		interval = nvram_get_int("upnp_clean_interval");
-		if (interval < 60)
-			interval = 60;
-
-		fprintf(f, "clean_ruleset_interval=%d\n"
-		           "clean_ruleset_threshold=%d\n",
-		           interval,
-		           nvram_get_int("upnp_clean_threshold"));
-	}
-	else
-		fprintf(f, "clean_ruleset_interval=0\n");
-
 	https = nvram_get_int("https_enable");
 	fprintf(f, "presentation_url=http%s://%s:%s/forward-upnp.asp\n", (https ? "s" : ""), nvram_safe_get("lan_ipaddr"), nvram_safe_get(https ? "https_lanport" : "http_lanport"));
 
--- a/release/src-rt-6.x.4708/router/shared/defaults.c
+++ b/release/src-rt-6.x.4708/router/shared/defaults.c
@@ -85,7 +85,6 @@ struct nvram_tuple upnp_defaults[] = {
 	{ "upnp_secure",		"1"				, 0 },
 	{ "upnp_port",			"0"				, 0 },
 	{ "upnp_ssdp_interval",		"900"				, 0 },	/* SSDP interval */
-	{ "upnp_mnp",			"0"				, 0 },
 	{ "upnp_custom",		""				, 0 },
 	{ "upnp_lan",			""				, 0 },
 	{ "upnp_lan1",			""				, 0 },
--- a/release/src-rt-6.x.4708/router/shared/defaults.c
+++ b/release/src-rt-6.x.4708/router/shared/defaults.c
@@ -90,9 +90,6 @@ struct nvram_tuple upnp_defaults[] = {
 	{ "upnp_lan1",			""				, 0 },
 	{ "upnp_lan2",			""				, 0 },
 	{ "upnp_lan3",			""				, 0 },
-	{ "upnp_clean",			"1"				, 0 },	/* 0:Disable 1:Enable */
-	{ "upnp_clean_interval",	"600"				, 0 },	/* Cleaning interval in seconds */
-	{ "upnp_clean_threshold",	"20"				, 0 },	/* Threshold for cleaning unused rules */
 #if 0	/* disabled for miniupnpd */
 	{ "upnp_max_age",		"180"				, 0 },	/* Max age */
 	{ "upnp_config",		"0"				, 0 },
--- a/release/src-rt-6.x.4708/router/www/forward-upnp.asp
+++ b/release/src-rt-6.x.4708/router/www/forward-upnp.asp
@@ -11,7 +11,7 @@
 <head>
 <meta http-equiv="content-type" content="text/html;charset=utf-8">
 <meta name="robots" content="noindex,nofollow">
-<title>[<% ident(); %>] Forwarding: UPnP / NAT-PMP</title>
+<title>[<% ident(); %>] Forwarding: UPnP IGD &amp; PCP/NAT-PMP</title>
 <link rel="stylesheet" type="text/css" href="tomato.css?rel=<% version(); %>">
 <% css(); %>
 <script src="isup.jsz?rel=<% version(); %>"></script>
@@ -49,11 +49,11 @@ function upnpNvramAdd() {
 		return;
 
 	E('_f_enable_upnp').disabled = 1;
-	E('_f_enable_natpmp').disabled = 1;
+	E('_f_enable_pcp_pmp').disabled = 1;
 	if ((sb = E('save-button')) != null) sb.disabled = 1;
 	if ((cb = E('cancel-button')) != null) cb.disabled = 1;
 
-	if (!confirm("Add UPNP to nvram?"))
+	if (!confirm("Add UPnP IGD & PCP/NAT-PMP autonomous port forwarding service to nvram?"))
 		return;
 
 	if (xob)
@@ -76,7 +76,7 @@ function upnpNvramAdd() {
 		setTimeout(
 			function() {
 				E('_f_enable_upnp').disabled = 0;
-				E('_f_enable_natpmp').disabled = 0;
+				E('_f_enable_pcp_pmp').disabled = 0;
 				if (sb) sb.disabled = 0;
 				if (cb) cb.disabled = 0;
 				if (msg) msg.style.display = 'none';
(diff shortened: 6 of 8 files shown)


pedro0311:
Oh, I'm sorry I do not monitor this mirror of FreshTomato ARM.
Also I can't merge it here, so unfortunately you have to do this PR on bitbucket: https://bitbucket.org/pedro311/freshtomato-arm/pull-requests/new

BR

pedro0311:
I can import it manually to my local repository and then use it (push it to bitbucket), but please combine it to only one commit.

BR

Self-Hosting-Group:
A few suggestions that have come up:

- The secure mode configuration only applies to UPnP IGD, for PCP there is an additional configuration option `pcp_allow_thirdparty` in the daemon. How about making the option in the UI also apply to PCP, possibly inverting the option and rewording it as it is currently pretty much always enabled, while keeping `nvram.upnp_secure`?
- Enabling the service with NVRAM removal requires at least 4 clicks. How about reducing this by checking the UPnP IGD and PCP/NAT-PMP protocols immediately after adding NVRAM, or having a UI option to enable the service/NVRAM and have the protocols enabled by default?

pedro0311:
> The secure mode configuration only applies to UPnP IGD, for PCP there is an additional configuration option pcp_allow_thirdparty in the daemon. How about making the option in the UI also apply to PCP, possibly inverting the option and rewording it as it is currently pretty much always enabled, while keeping nvram.upnp_secure?

Good idea, just name it something to upnp_secure_pcp (so inverted)

> Enabling the service with NVRAM removal requires at least 4 clicks. How about reducing this by checking the UPnP IGD and PCP/NAT-PMP protocols immediately after adding NVRAM, or having a UI option to enable the service/NVRAM and have the protocols enabled by default?

Better not. This would require changes to all GUI pages that use the method of removing nvram entries when they are not being used.

BR

Self-Hosting-Group:
Enable protocol options here?
https://github.com/pedro0311/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/www/forward-upnp.asp#L51-L52

Regarding the removal of the `Inactive Rules Cleaning` options. It seems to work partially, some port maps are removed, some are not. Upstream issue created. But the functionality is IMHO against the expectation of the clients, as valid unexpired port maps are removed if there is no traffic. So inactive means no traffic! The cleanup of expired port maps works independently and works. What to do now?
1. Remove the UI options anyway, because it's not standard
2. Keep it for now, but rewrite it so it's clearyfied
3. Same as 2, but disable it by default

pedro0311:
> Enable protocol options here?

You were right, let's do as you said:

> How about making the option in the UI also apply to PCP, possibly inverting the option and rewording it as it is currently pretty much always enabled, while keeping nvram.upnp_secure?

And:
> What to do now?
> Remove the UI options anyway, because it's not standard

Right, sometimes less is more.

pedro0311:
I'll check it when I have time, thanks.

pedro0311:
OK, I've got it working.
Question though: why compile with igd2 and disable at runtime? The only change is that miniupnpd is bigger.

pedro0311:
Added: https://bitbucket.org/pedro311/freshtomato-arm/commits/0ce92cb30c3de66a9d2b4158536572f066627245

Self-Hosting-Group:
Wow, that was fast, almost too fast, I might change `f_upnp_secure` -> `f_allow_third_party` to make it clearer. And as for IGDv2, yes, there is a reason for (infinite) port maps to be removed when compiling with, notes to follow. Have you checked why the IPv6 chain is not created, for use with e.g. PCP? I just noticed that the protocols do not seem to be checked after nvram.

pedro0311:
> I might change f_upnp_secure -> f_allow_third_party to make it clearer.

It doesn't matter, it's only local variable.

> And as for IGDv2, yes, there is a reason for (infinite) port maps to be removed when compiling with, notes to follow.

Can you elaborate this?

> Have you checked why the IPv6 chain is not created, for use with e.g. PCP?

I don't know, but it always was like this (even 14 years ago, on tomato by shibby). I already asked @MarsTomato if he has time and can look into it (I don't have access to IPv6).

BR

Self-Hosting-Group:
After running `ip6tables -t filter -N MINIUPNPD` the daemon stops complaining and the port map is created. Perhaps it should be added here: https://github.com/pedro0311/freshtomato-arm/blob/0ce92cb30c3de66a9d2b4158536572f066627245/release/src-rt-6.x.4708/router/rc/firewall.c#L945-L989

MarsTomato:
> After running `ip6tables -t filter -N MINIUPNPD` the daemon stops complaining and the port map is created. Perhaps it should be added here:

Can you show me what you get / see? I do not see any complaints so far? (R7000 VPN Version 2024.3 release)

```
root@R7000:/tmp# miniupnpd -f config.alt -d
miniupnpd[10886]: version 2.3.7 started
miniupnpd[10886]: system uptime is 4099345 seconds
miniupnpd[10886]: Reloading rules from lease file
miniupnpd[10886]: version 2.3.7 starting NAT-PMP/PCP UPnP-IGD ext if ppp0 BOOTID=1732639534
miniupnpd[10886]: HTTP listening on port 30166
miniupnpd[10886]: HTTP IPv6 address given to control points : [2003:d4aa:3fbb:aa00::1]
miniupnpd[10886]: Listening for NAT-PMP/PCP traffic on port 5351
miniupnpd[10886]: HTTP REQUEST from [::ffff:10.xx.yy.100]:62938 : GET /rootDesc.xml (HTTP/1.1)
miniupnpd[10886]: Host: 10.xx.yy.1:30166
miniupnpd[10886]: level=0 type=8
miniupnpd[10886]: ifindex = 12  10.xx.yy.1
miniupnpd[10886]: ST: urn:dial-multiscreen-org:service:dial:1 (ver=1)
miniupnpd[10886]: SSDP M-SEARCH from 10.xx.yy.148:57720 ST: urn:dial-multiscreen-org:service:dial:1
miniupnpd[10886]: Received UDP Packet (IPv6)
miniupnpd[10886]: ST: urn:dial-multiscreen-org:service:dial:1 (ver=1)
miniupnpd[10886]: SSDP M-SEARCH from [::ffff:10.xx.yy.148]:57720 ST: urn:dial-multiscreen-org:service:dial:1
miniupnpd[10886]: level=0 type=8
miniupnpd[10886]: ifindex = 12  10.xx.yy.1
miniupnpd[10886]: ST: urn:dial-multiscreen-org:service:dial:1 (ver=1)
miniupnpd[10886]: SSDP M-SEARCH from 10.xx.yy.148:51808 ST: urn:dial-multiscreen-org:service:dial:1
miniupnpd[10886]: Received UDP Packet (IPv6)
miniupnpd[10886]: ST: urn:dial-multiscreen-org:service:dial:1 (ver=1)
miniupnpd[10886]: SSDP M-SEARCH from [::ffff:10.xx.yy.148]:51808 ST: urn:dial-multiscreen-org:service:dial:1
miniupnpd[10886]: level=0 type=8
miniupnpd[10886]: ifindex = 12  10.xx.yy.1
miniupnpd[10886]: ST: upnp:rootdevice (ver=0)
miniupnpd[10886]: SSDP M-SEARCH from 10.xx.yy.100:50289 ST: upnp:rootdevice
miniupnpd[10886]: Single search found
miniupnpd[10886]: SendSSDPResponse(): 0 bytes to 10.xx.yy.100:50289 ST: HTTP/1.1 200 OK
CACHE-CONTROL: max-age=1800
ST: upnp:rootdevice
USN: uuid:a96e8b84-dc04-4512-815b-3584794ee8e7::upnp:rootdevice
EXT:
SERVER: FreshTomato/2.6.36.4brcmarm UPnP/1.1 MiniUPnPd/2.3.7
LOCATION: http://10.xx.yy.1:30166/rootDesc.xml
OPT: "http://schemas.upnp.org/upnp/1/0/"; ns=01
01-NLS: 1732639534
BOOTID.UPNP.ORG: 1732639534
CONFIGID.UPNP.ORG: 1337


miniupnpd[10886]: Received UDP Packet (IPv6)
miniupnpd[10886]: get_lan_for_peer() looking for LAN interface index=12
miniupnpd[10886]: ifname=br0 index=12 str=10.xx.yy.1 addr=0a0a3201 mask=ffffff00
miniupnpd[10886]: ST: upnp:rootdevice (ver=0)
miniupnpd[10886]: SSDP M-SEARCH from [fe80::dffd:b233:325c:ae30%br0]:50286 ST: upnp:rootdevice
miniupnpd[10886]: Single search found
miniupnpd[10886]: SendSSDPResponse(): 0 bytes to [fe80::dffd:b233:325c:ae30%br0]:50286 ST: HTTP/1.1 200 OK
CACHE-CONTROL: max-age=1800
ST: upnp:rootdevice
USN: uuid:a96e8b84-dc04-4512-815b-3584794ee8e7::upnp:rootdevice
EXT:
SERVER: FreshTomato/2.6.36.4brcmarm UPnP/1.1 MiniUPnPd/2.3.7
LOCATION: http://[2003:d4aa:3fbb:aa00::1]:30166/rootDesc.xml
OPT: "http://schemas.upnp.org/upnp/1/0/"; ns=01
01-NLS: 1732639534
BOOTID.UPNP.ORG: 1732639534
CONFIGID.UPNP.ORG: 1337


miniupnpd[10886]: Received UDP Packet (IPv6)
```
...

Self-Hosting-Group:
Have you successfully created an IPv6 port map using PCP? I get the following log when I try to do this:

unknown daemon.err miniupnpd ...: chain MINIUPNPD not found
unknown daemon.err miniupnpd ...: PCP MAP: failed to add mapping ...

After viewing the initialisation script, I think the following commands must be executed:

ip6tables -t filter -N MINIUPNPD
ip6tables -t filter -I FORWARD -i $EXTIF ! -o $EXTIF -j MINIUPNPD

See: https://github.com/miniupnp/miniupnp/blob/master/miniupnpd/netfilter/ip6tables_init.sh

Edit: sorry, restore your comment

MarsTomato:
Can you provide more details? custom config? You are using a custom build/image with "--igd2" support? No IPv6 Pinhole support without igd2 if i remember right?

Self-Hosting-Group:
> Can you provide more details? custom config? You are using a custom build/image with "--igd2" support? No IPv6 Pinhole support without igd2 if i remember right?

Standard image, no custom configuration, and no. UPnP IGD and PCP are two completely different protocols, as daemon is built with `configure --ipv6` it has IPv6 support for PCP, for IGD it needs to be compiled with `--igd2` in addition.

MarsTomato:
> > Can you provide more details? custom config? You are using a custom build/image with "--igd2" support? No IPv6 Pinhole support without igd2 if i remember right?
> 
> Standard image, no custom configuration, and no. UPnP IGD and PCP are two completely different protocols, as daemon is built with `configure --ipv6` it has IPv6 support for PCP, for IGD it needs to be compiled with `--igd2` in addition.

Yes, right. IGDv1 is too old for IPv6 :)

Does RMerlin have a chain in place for IPv6? I do not see it right now ... mhh.
Can you please tell me what Client Software / App do you use to test PCP IPv6 ? something i can use to provoke the miniupnd warning
OR
does this warning come right away after starting miniupnpd?

Self-Hosting-Group:
> Does RMerlin have a chain in place for IPv6? I do not see it right now ... mhh.

Have you explicitly enabled IPv6/IGDv2 support in the UI? I have not tested IPv6 support there.

> Can you please tell me what Client Software App do you use to test PCP IPv6 ? something i can use to provoke the miniupnd warning

https://github.com/libpcp/pcp/tree/master/pcp_app

MarsTomato:
> > Does RMerlin have a chain in place for IPv6? I do not see it right now ... mhh.
> 
> Have you explicitly enabled IPv6/IGDv2 support in the UI? I have not tested IPv6 support there.
> 
> > Can you please tell me what Client Software App do you use to test PCP IPv6 ? something i can use to provoke the miniupnd warning
> 
> https://github.com/libpcp/pcp/tree/master/pcp_app

I have IPv6 enabled but FreshTomato Version 2024.3 is build with IGDv1. 

`# Disable IPv6 (default is no : IPv6 enabled if enabled at build time)
#ipv6_disable=yes`

see also syslog above

`miniupnpd[10886]: HTTP IPv6 address given to control points : [2003:d4aa:3fbb:aa00::1]
miniupnpd[10886]: Listening for NAT-PMP/PCP traffic on port 5351`

`miniupnpd[10886]: Received UDP Packet (IPv6)
miniupnpd[10886]: get_lan_for_peer() looking for LAN interface index=12`


Do you see this miniupnpd warning right away after starting miniupnpd?

Does RMerlin have an IPv6 Chain in place? I see he added "--igd2"



> https://github.com/libpcp/pcp/tree/master/pcp_app

I am more looking for an App/Software to make use of this library

Self-Hosting-Group:
We may have just discovered an upstream bug, another thing where the daemon behaves inconsistently is that it uses the `MINIUPNPD` ip6tables chain name and not the configured `upnp` as with IPv4, which is set in `/etc/upnp/config`. As the daemon is mostly used with nftables, this may have gone unnoticed for such a long time. The ip6tables chain name is defined in the following line and does not seem to be overridden by the configuration file later in the code, as with IPv4 and the nftables and pf backends:

https://github.com/miniupnp/miniupnp/blob/0ff36e5397f36cd0bda626bc548192d10832d57f/miniupnpd/netfilter/iptpinhole.c#L31

Maintainer ping @miniupnp

Self-Hosting-Group:
> I am more looking for an App/Software to make use of this library

None known to me, I had to compile it myself to test and see the behaviour.

> I have IPv6 enabled but FreshTomato Version 2024.3 is build with IGDv1.

I noticed this. See my suggested commit in the PR: https://github.com/pedro0311/freshtomato-arm/pull/4/commits/e1e11ef6d38d628c02e84bcc9692a5eb6b1a5aff.

Self-Hosting-Group:
But if you don't want to compile it yourself, I can offer you to check the proposed changes if you provide me with a test image.

MarsTomato:
> We may have just discovered an upstream bug, another thing where the daemon behaves inconsistently is that it uses the `MINIUPNPD` ip6tables chain name and not the configured `upnp` as with IPv4, which is set in `/etc/upnp/config`. As the daemon is mostly used with nftables, this may have gone unnoticed for such a long time. The ip6tables chain name is defined in the following line and does not seem to be overridden by the configuration file later in the code, as with IPv4 and the nftables and pf backends:
> 
> https://github.com/miniupnp/miniupnp/blob/0ff36e5397f36cd0bda626bc548192d10832d57f/miniupnpd/netfilter/iptpinhole.c#L31
> 
> Maintainer ping @miniupnp

Yes, i also was wondering why we have
`upnp_forward_chain=MyNameXYZ`
`upnp_nat_chain=MyNameXYZ`
`upnp_nat_postrouting_chain=MyNameXYZ`
for IPv4 but no option for IPv6 at all?

mmhh -  ip6tables chain name should be always "MINIUPNPD" (at least right now)

MarsTomato:
> > I am more looking for an App/Software to make use of this library
> 
> None known to me, I had to compile it myself to test and see the behaviour.
> 
> > I have IPv6 enabled but FreshTomato Version 2024.3 is build with IGDv1.
> 
> I noticed this. See my suggested commit in the PR: [e1e11ef](https://github.com/pedro0311/freshtomato-arm/commit/e1e11ef6d38d628c02e84bcc9692a5eb6b1a5aff).

IGDv2 it is already partly in place (at least the comment) :)

MarsTomato:
> But if you don't want to compile it yourself, I can offer you to check the proposed changes if you provide me with a test image.

What image do you need? Router? I prepared the changes and have it already running on my side.

Self-Hosting-Group:
> What image do you need? Router? I prepared the changes and have it already running on my side.

Thanks for the offer, I'm testing with a RT-AC68U router. I noticed that the daemon sometimes starts too early when IPv6 is not yet usable, and then IPv6 is disabled in the daemon. This may be because IPv6 only works here if I enable `Request PD Only` with freshtomato, which may take longer than SLAAC. Perhaps a certain uptime should be required before the daemon starts. Also, it would be great if the protocols were automatically enabled after adding them to the nvram (LANs still need to be selected), I think the following commit should do that f6e58f2.

Daemon log starting too early:

```
user	info	ntpd_synced[1088]	initial clock set
daemon	notice	miniupnpd[1090]	version 2.3.7 started
daemon	warn	miniupnpd[1090]	no HTTP IPv6 address, disabling IPv6
```

Commit 9e4e3ce and changelog for 2018.4:

```
router/rc/wan.c: start miniupnpd after httpd/later to avoid disabling IPv6 at miniupnpd startup (does happen sometimes with 2018.3, solves miniupnpd warning "no HTTP IPv6 address, disabling IPv6" at reboot/restart)
```

It might be a good idea to use the latest untagged daemon code, the configure/makefiles have been heavily revised since the last release and may not need some of the following patches:
https://github.com/pedro0311/freshtomato-arm/tree/arm-master/release/src-rt-6.x.4708/router/patches/miniupnpd

I'm preparing a PR with further improvements, but I still want to test it with the image/newly configured daemon and I was wondering if you know if there is already a function in freshtomato that generates the nice duration string like the uptime on the main page?

Self-Hosting-Group:
The too early daemon start is not always the case, so I have revised the comment.

MarsTomato:
> > What image do you need? Router? I prepared the changes and have it already running on my side.
> 
> Thanks for the offer, I'm testing with a RT-AC68U router. I noticed that the daemon sometimes starts too early when IPv6 is not yet usable, and then IPv6 is disabled in the daemon. This may be because IPv6 only works here if I enable `Request PD Only` with freshtomato, which may take longer than SLAAC. 

We need a prefix for br0 (so you need to enable "Request PD Only")
`# It is mandatory to use the network interface name in order to enable IPv6`
`listening_ip=br0`

I uploaded almost latest branch images, see
https://bitbucket.org/M_ars/freshtomato-arm/downloads/

(Currently) We start miniupnd after WAN up (IPv4 only). Usually IPv6 comes right away, sometimes its the first one. In 2018.4 WAN-up (IPv4) and IPv6 was glued together, now it is independet

Self-Hosting-Group:
> I uploaded almost latest branch images, see https://bitbucket.org/M_ars/freshtomato-arm/downloads/

Thanks for the image, now the IPv6 chain is created and port maps work with PCP and IPv6, thanks also for including IGDv2 which is disabled at runtime, I will mention later in the upcoming PR why this is good.

> (Currently) We start miniupnd after WAN up (IPv4 only). Usually IPv6 comes right away, sometimes its the first one. In 2018.4 WAN-up (IPv4) and IPv6 was glued together, now it is independet

Now I have also tested with a connection where IPv6 works without `Request PD Only`, again the daemon starts too early most of the time with the device. Since the daemon does not handle the not yet available address itself, and IPv6 is then permanently disabled, we could perhaps simply delay the start of the daemon by an additional time (possibly 20s), or preferably additionally wait for IPv6 with such a timeout. According to the log, IPv6 is ready a few seconds (~2s) after the daemon is started, so less than 20s should be enough.

FreshTomato uses special (probably early) code for lease file support in the daemon, with a small bug with long port forward descriptions. However, in the long term, the simpler standard `lease_file` configuration option could be used instead, and the special code removed from the daemon. A port forward can then be removed by removing the line in the file and restarting the daemon, as e.g. OpenWrt does.

https://github.com/openwrt/luci/blob/master/applications/luci-app-upnp/root/usr/share/rpcd/ucode/luci.upnp#L115-L134

MarsTomato:
> Now I have also tested with a connection where IPv6 works without `Request PD Only`, again the daemon starts too early most of the time with the device. Since the daemon does not handle the not yet available address itself, and IPv6 is then permanently disabled, we could perhaps simply delay the start of the daemon by an additional time (possibly 20s), or preferably additionally wait for IPv6 with such a timeout. According to the log, IPv6 is ready a few seconds (~2s) after the daemon is started, so less than 20s should be enough.
> 
adjusted it (restart deamon with IPv6 address update or new one)
please test again with "04122024v001" images
https://bitbucket.org/M_ars/freshtomato-arm/downloads/

> FreshTomato uses special (probably early) code for lease file support in the daemon, with a small bug with long port forward descriptions. However, in the long term, the simpler standard `lease_file` configuration option could be used instead, and the special code removed from the daemon. A port forward can then be removed by removing the line in the file and restarting the daemon, as e.g. OpenWrt does.
> 

Yes, i can see it too. "lease_file" option should be the better way for the future.

Self-Hosting-Group:
Thank you very much. Now the daemon starts reliably, even after many reboot tests. However, in my case it always starts twice in quick succession. Each time the daemon is started, all UPnP devices and services must first be de-announced via SSDP multicast (to clear any previous announcements), then announced, and then de-announced again on exit, and this now happens twice. The second daemon start is ~2s later. Maybe this would not be necessary if we waited a bit, also because I think that a later start of the service is not a limitation. But I am happy with the current configuration and would like to thank you both again for the great teamwork. PS: Small PR on service UI improvements coming next week.

PR #5, opened 2024-12-18
UPnP IGD & PCP: Add `Expires` to port forward listing

Self-Hosting-Group:
- Add `Expires` to port forward listing
- Improve UI and wording
- Compile daemon with IGDv2 (`--igd2`) by default for all builds, but disable it by default at runtime to workaround IGDv2 incompatible clients, but have the benefit of limit infinite port maps to a maximum of 7d and return more accurate errors
- Compile daemon with `--disable-pppconn` as legacy workaround

freshtomato
_[Screenshot 2024.3 release](https://github.com/user-attachments/assets/57e42a93-38f2-49f1-afd1-b3418ea453c6)_

The **Port Control Protocol (PCP)** is the successor to NAT-PMP, has similar protocol concepts and packet formats, but adds support for **IPv6 port maps** and options/extensions. For more information, see:
**Port Mapping Protocols Overview and Comparison 2024: About UPnP IGD & PCP/NAT-PMP**
https://github.com/Self-Hosting-Group/wiki/wiki/Port-Mapping-Protocols-Overview

Previous improvement: #4

### DIFF
--- a/release/src-rt-6.x.4708/router/Makefile
+++ b/release/src-rt-6.x.4708/router/Makefile
@@ -1519,7 +1519,7 @@ miniupnpd/stamp-h1:
 	$(call patch_files,miniupnpd)
 	cd miniupnpd && \
 		PKG_CONFIG_LIBDIR="" \
-		./configure --leasefile --vendorcfg --portinuse --firewall=iptables --iptablespath=$(TOP)/$(IPTABLES_TARGET) $(if $(TCONFIG_IPV6),--ipv6 --igd2,)
+		./configure --igd2 --leasefile --vendorcfg --portinuse --disable-pppconn --firewall=iptables --iptablespath=$(TOP)/$(IPTABLES_TARGET) $(if $(TCONFIG_IPV6),--ipv6,)
 	@touch $@
 
 miniupnpd: miniupnpd/stamp-h1
--- a/release/src-rt-6.x.4708/router/rc/services.c
+++ b/release/src-rt-6.x.4708/router/rc/services.c
@@ -1819,9 +1819,7 @@ void start_upnp(void)
 	           "port=%d\n"
 	           "enable_upnp=%s\n"
 	           "enable_pcp_pmp=%s\n"
-#ifdef TCONFIG_IPV6
-	           "force_igd_desc_v1=yes\n" /* If igd2 is compiled in + IPv6 support (force version 1 in IGD v2 mode) */
-#endif
+	           "force_igd_desc_v1=yes\n"
 	           "secure_mode=%s\n"
 	           "pcp_allow_thirdparty=%s\n"
 	           "upnp_forward_chain=upnp\n"
@@ -1834,6 +1832,9 @@ void start_upnp(void)
 	           "model_url=https://freshtomato.org/\n"
 	           "manufacturer_name=FreshTomato Firmware\n"
 	           "manufacturer_url=https://freshtomato.org/\n"
+	           /* Empty strings so that 1 and 00000000 are not reported */
+	           "model_number=\n"
+	           "serial=\n"
 	           "\n",
 	           get_wanface("wan"),
 	           upnp_port,
--- a/release/src-rt-6.x.4708/router/www/forward-upnp.asp
+++ b/release/src-rt-6.x.4708/router/www/forward-upnp.asp
@@ -48,7 +48,7 @@ function upnpNvramAdd() {
 	if ((sb = E('save-button')) != null) sb.disabled = 1;
 	if ((cb = E('cancel-button')) != null) cb.disabled = 1;
 
-	if (!confirm("Add UPnP IGD & PCP/NAT-PMP autonomous port forwarding service to nvram?"))
+	if (!confirm("Add UPnP IGD & PCP/NAT-PMP autonomous port forwarding service to NVRAM?"))
 		return;
 
 	if (xob)
@@ -132,7 +132,7 @@ function submitDelete(proto, eport) {
 }
 
 function deleteData(data) {
-	if (!confirm('Delete port forward '+data[2]+'/'+data[3]+' -> '+data[0]+':'+data[1]+'?'))
+	if (!confirm('Delete port forward for '+data[0]+':'+data[1]+'/'+data[3]+' ('+data[5]+')?'))
 		return;
 
 	submitDelete(data[3], data[2]);
@@ -145,11 +145,17 @@ function deleteAll() {
 	submitDelete('*', '0');
 }
 
+function padnum(num, length) {
+	num = num.toString();
+	while (num.length < length) num = "0" + num;
+	return num;
+}
+
 var ug = new TomatoGrid();
 
 ug.setup = function() {
 	this.init('upnp-grid', 'sort delete');
-	this.headerSet(['Internal Address', 'Internal Port', 'External Port', 'Protocol', 'Description']);
+	this.headerSet(['Client Address', 'Client Port', 'Ext Port', 'Protocol', 'Expires', 'Description']);
 	ug.populate();
 	this.sort(0);
 }
@@ -164,13 +170,27 @@ ug.populate = function() {
 			if (r == null)
 				continue;
 
-			row = this.insertData(-1, [r[3], r[4], r[2], r[1], r[5]]);
+			var expires_sec, hour, minute, second, expires_str = '';
+			expires_sec = (new Date(r[6] * 1000) - new Date().getTime()) / 1000;
+			hour = Math.floor(expires_sec / 3600);
+			minute = Math.floor(expires_sec % 3600 / 60);
+			second = Math.floor(expires_sec % 60);
+			if (hour > 0) {
+				expires_str += hour + 'h ';
+				expires_str += padnum(minute, 2) + 'm ';
+				expires_str += padnum(second, 2) + 's';
+			} else if (minute > 0) {
+				expires_str += minute + 'm ';
+				expires_str += padnum(second, 2) + 's';
+			} else if (expires_sec > 0) expires_str += second + 's';
+
+			row = this.insertData(-1, [r[3], r[4], r[2], r[1], expires_str, r[5]]);
 
 			if (!r[0]) {
-				for (j = 0; j < 5; ++j)
+				for (j = 0; j < 6; ++j)
 					elem.addClass(row.cells[j], 'disabled');
 			}
-			for (j = 0; j < 5; ++j)
+			for (j = 0; j < 6; ++j)
 				row.cells[j].title = 'Delete';
 		}
 	}
@@ -188,7 +208,7 @@ ug.onClick = function(cell) {
 function verifyFields(focused, quiet) {
 	var enable = (E('_f_enable_upnp').checked || E('_f_enable_pcp_pmp').checked);
 
-	E('_f_upnp_secure').disabled = !enable;
+	E('_f_upnp_allow_third_party').disabled = !enable;
 	E('_upnp_custom').disabled = !enable;
 
 	for (var i = 0 ; i <= MAX_BRIDGE_ID ; i++) {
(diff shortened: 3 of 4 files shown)


Self-Hosting-Group:
Draft, as it would be nice to have the UPnP IGD & PCP/NAT-PMP protocols enabled (without LANs) once the variables have been added to NVRAM, as this is probably what the user wants, but I couldn't get where best to define this.

Since I suggested compiling UPnP IGDv2 in miniupnpd and disabling it at runtime (for compatibility reasons), I would like to explain the differences of this change:

1. the daemon now reduces infinite/static (value `0`) port forwards (only valid with IGDv1, deprecated) to the maximum duration of 7d according to the IGDv2 standard, this will delete infinite port forwards after this time that have not been deleted by legacy clients
2. the daemon now uses the extended IGDv2 error codes, e.g. deny by ACL with `Action not authorised` instead of `ConflictInMappingEntry`
3. the daemon now always writes the UNIX epoc to the lease file and no longer a `0` for infinite port forwards, so you can always get the correct expires ;-)

(Note: this applies if IGDv2 is compiled into the daemon and IGDv1 is enabled at runtime, NOT if only IGDv1 is compiled in)

The time calculation is done on the client, the backend would probably be more accurate.

I chose `expires` because it is clear and short. However, another term could be used, the following come to mind:

Port forward ...
- valid until
- expiration
- time left
- duration
- lifetime
- expires
- expiry
- valid

MarsTomato:
good idea :)
Can/could you also check one of the advanded themes, for example "Advanced Red" ?

Non-IPv6 FreshTomato images are build currently with IGDv1 only, so "leaseduration" will be 0 .
So we may need to make your GUI changes conditional to IPv6 (TCONFIG_IPV6) ?

OR include IGDv2 always and adjust the config file ... mhhh
https://bitbucket.org/pedro311/freshtomato-arm/commits/bb80103323484451db37ac90d4bbcd99ea5ca6cf

https://bitbucket.org/pedro311/freshtomato-arm/commits/bb80103323484451db37ac90d4bbcd99ea5ca6cf#Lrelease/src-rt-6.x.4708/router/rc/services.cT1823

@pedro0311 What do you think?

Self-Hosting-Group:
> good idea :) Can/could you also check one of the advanded themes, for example "Advanced Red" ?

Just checked, looks good.
 
> Non-IPv6 FreshTomato images are build currently with IGDv1 only, so "leaseduration" will be 0 . So we may need to make your GUI changes conditional to IPv6 (TCONFIG_IPV6) ?

Non-IPv6 FreshTomato images, can you elaborate on this, these are for older routers, is there a list of which ones are like this? Even with only IGDv1 compiled, most port maps are not `0`/infinite as supported by IGDv1 clients. The expiration time calculation I've created is also designed to not show anything for a `0`, which would result in a negative value.

> OR include IGDv2 always and adjust the config file ... mhhh https://bitbucket.org/pedro311/freshtomato-arm/commits/bb80103323484451db37ac90d4bbcd99ea5ca6cf
> 
> https://bitbucket.org/pedro311/freshtomato-arm/commits/bb80103323484451db37ac90d4bbcd99ea5ca6cf#Lrelease/src-rt-6.x.4708/router/rc/services.cT1823
> 
> @pedro0311 What do you think?

If IGDv2 is always compiled in, even for non-IPv6 builds, this has the advantage of having the same behaviour everywhere, e.g. limiting IPv4 port maps to 7d, which I think is good. If we accept the 4kB size difference. OpenWrt builds have been built this way for years (disabled at runtime by default).

MarsTomato:
> If IGDv2 is always compiled in, even for non-IPv6 builds, this has the advantage of having the same behaviour everywhere, e.g. limiting IPv4 port maps to 7d, which I think is good. If we accept the 4kB size difference. OpenWrt builds have been built this way for years (disabled at runtime by default).

Yes, i agree.

Probably best to extend this commit a little to include IGDv2 always now (+config file)

> Just checked, looks good.

perfect, thx 👍 

lets wait for @pedro0311  what is the easiest way for him to add (&extend) it. He is busy the next few hours probably :)

pedro0311:
> If we accept the 4kB size difference.

Let's compile all targets with --igd2 even for "non IPv6" ones.

Self-Hosting-Group:
Self-measured, I see a ~8kB difference when building the daemon for x86-64 with iptables outside of FreshTomato.

Self-Hosting-Group:
What do you think of this new commit? I also disabled `WANPPPConnection` with `--disable-pppconn` (option is included with `--strict`) because this workaround is very old. None of the other implementations I have seen advertise this service as a fallback to `WANIPConnection`, which is now used. Not even a commercial router, I just tested one that is ~17 years old. Set `model_number`/`serial` to an empty string so that the daemon does not report `1` and `00000000` as default.

> Draft, as it would be nice to have the UPnP IGD & PCP/NAT-PMP protocols enabled (without LANs) once the variables have been added to NVRAM, as this is probably what the user wants, but I couldn't get where best to define this.

Good idea?

Self-Hosting-Group:
Can the PR be used for the 2024.5 release?

MarsTomato:
> What do you think of this new commit? I also disabled `WANPPPConnection` with `--disable-pppconn` (option is included with `--strict`) because this workaround is very old. None of the other implementations I have seen advertise this service as a fallback to `WANIPConnection`, which is now used. Not even a commercial router, I just tested one that is ~17 years old. Set `model_number`/`serial` to an empty string so that the daemon does not report `1` and `00000000` as default.
> 
> > Draft, as it would be nice to have the UPnP IGD & PCP/NAT-PMP protocols enabled (without LANs) once the variables have been added to NVRAM, as this is probably what the user wants, but I couldn't get where best to define this.
> 
> Good idea?

I will check it in few days. Currently busy with christmas

MarsTomato:
> Can the PR be used for the 2024.5 release?

sorry, next release :) 
Building process for 2024.5 release was started like ~ 4 days ago

pedro0311:
> I will check it in few days. Currently busy with christmas

Me too.

MarsTomato:
looks good 👍

I only have small suggestions/improvments for the commit content:
1) move the config changes ("--igd2" + "force_igd_desc_v1=yes\n") from your second commit to the first commit. It is needed for the leaseduration/lifetime
2) add some comment for model_number/serial to an empty string (C-file or commit msg)

OR put it all in one commit? (+some comments / notes for the changes)

Self-Hosting-Group:
I have merged the commits into one and added a comment. Actually it would be better if the actualy time calculation is done with the backend/router time as currently with this line in JS: `expires_sec = (new Date(r[6] * 1000) - new Date().getTime()) / 1000;`. Can we do this in the ASP page, but only if it's not too complicated?

MarsTomato:
thx :)

>  Actually it would be better if the actualy time calculation is done with the backend/router time

Yes, it would be better, but this improvement can be done later. (We could add a patch for miniupnp)

@pedro0311 commit looks good in my opinion

pedro0311:
Done! Thanks!

Self-Hosting-Group:
Could it be that the freshtomato-ax repository is behind arm/mips? I see missing commits there.

pedro0311:
AX is totally different story, so don't ask ;)

Self-Hosting-Group:
Thank you. I see now that the daemon version there is 2.2.0, maybe from Asus ;-)
I was just wondering if it is always (including IPv4-only) built with `--igd2` and configured in services.c with `force_igd_desc_v1=yes` as on the other platforms, or if you have to leave it as it is.

MarsTomato:
> Thank you. I see now that the daemon version there is 2.2.0, maybe from Asus ;-) I was just wondering if it is always (including IPv4-only) built with `--igd2` and configured in services.c with `force_igd_desc_v1=yes` as on the other platforms, or if you have to leave it as it is.

AX is not the same. But i am also thinking that we could come closer to arm/mips branch and adjust/improve AX branch also.

Self-Hosting-Group:
Thank you very much. I have another question about Tomato/FreshTomato and their special lease-file support in miniupnpd. If I were to change the web interface to use the general lease-file feature, could the code be removed from miniupnpd, or is there another x-Tomato project that would miss this code?

MarsTomato:
> Thank you very much. I have another question about Tomato/FreshTomato and their special lease-file support in miniupnpd. If I were to change the web interface to use the general lease-file feature, could the code be removed from miniupnpd, or is there another x-Tomato project that would miss this code?

FreshTomato is the only tomato project/branche i know/see right now. Last update for shibby-arm in 2017 ... so i think it should be possible to go that path

Self-Hosting-Group:
Hi @pedro0311 and @MarsTomato

I have a few ideas and questions about FreshTomato.

1. Are **wiki updates/improvements** of https://wiki.freshtomato.org/doku.php/forward-upnp welcome? Looks outdated, I can provide an updated document
2. I have a small **patch for miniupnpd**, this would mitigate the following issue https://redirect.github.com/FreshTomato-Project/freshtomato-arm/issues/75. It implements a dedicated workaround for the IGDv2 incompatible windows client. Would you like to apply it?
3. Set **default hostname** to `freshtomato` (for new setups in `defaults.c`), for easier access of router UI and default **domain name** to `internal` to mignate DNSSEC client validation issues ? See https://datatracker.ietf.org/doc/html/draft-davies-internal-tld-04
4. How about an **online demo UI** like DD-WRT and AsusWRT (e.g. demoui.freshtomato.org)? Can provide the script used at DD-WRT to grab the UI. As FreshTomato is also using HTTP basic auth, the script would not have a lot of adjustments, but when checked, the invoked tool could not automatically follow the navigation links, as these seems dynamically generated by JavaScript
5. Set the **moderner/responsive** `Advanced Red/Blue` (or black) **theme by default**?
6. **Automatically switch** `Advanced Red` to `Advanced Dark` **based on OS/browser** (UI option?), e.g. by CSS `prefers-color-scheme: dark` property, as OpenWrt does
7. **Increase** `Advanced *` theme header version (class `version`) text **contrast**, by e.g. using `color: #fff;` (or black for even more contrast).

Do you think these suggestions make sense or not?

HommeOursPorc:
> Hi @pedro0311 and @MarsTomato
> 
> I have a few ideas and questions about FreshTomato.
> 
>     1. Are **wiki updates/improvements** of https://wiki.freshtomato.org/doku.php/forward-upnp welcome? Looks outdated, I can provide an updated document
> 
Yes sure 😊, you'll need an account to submit changes. Check on the forum, maybe ask Hogwild for this.

>     4. How about an **online demo UI** like DD-WRT and AsusWRT (e.g. demoui.freshtomato.org)? Can provide the script used at DD-WRT to grab the UI. As FreshTomato is also using HTTP basic auth, the script would not have a lot of adjustments, but when checked, the invoked tool could not automatically follow the navigation links, as these seems dynamically generated by JavaScript
> 
This was discussed here: 
https://www.linksysinfo.org/index.php?threads/promoting-tomato.79071/post-357168

I don't know if @wellloaded did something with it finally??

MarsTomato:
> 2\. I have a small **patch for miniupnpd**, this would mitigate the following issue https://redirect.github.com/FreshTomato-Project/freshtomato-arm/issues/75. It implements a dedicated workaround for the IGDv2 incompatible windows client. Would you like to apply it?

Could you show us what you have in mind?

thx & BR

Self-Hosting-Group:
> Could you show us what you have in mind?

Sure. Can also create a PR.

```
--- a/upnpsoap.c
+++ b/upnpsoap.c
@@ -853,6 +853,9 @@ GetSpecificPortMappingEntry(struct upnphttp * h, const char * action, const char
 		       (unsigned int)iport, desc, leaseduration);
 #ifdef ENABLE_PCP
 		hide_pcp_nonce(desc);
+#endif
+#ifdef IGD_V2
+		if (h->respflags & FLAG_MS_CLIENT) leaseduration = 0;
 #endif
 		bodylen = snprintf(body, sizeof(body), resp,
 				action, ns/*SERVICE_TYPE_WANIPC*/,
@@ -1123,6 +1126,9 @@ GetGenericPortMappingEntry(struct upnphttp * h, const char * action, const char
 		char body[2048];
 #ifdef ENABLE_PCP
 		hide_pcp_nonce(desc);
+#endif
+#ifdef IGD_V2
+		if (h->respflags & FLAG_MS_CLIENT) leaseduration = 0;
 #endif
 		bodylen = snprintf(body, sizeof(body), resp,
 			action, ns, /*SERVICE_TYPE_WANIPC,*/ rhost,
```

The patch applies (if unchanged) to daemon version 2.3.9.

PR #54, opened 2025-01-29
GUI: admin-upgrade.asp - allow full filename display

wellloaded:
This change allows displaying the full file name (as distributed by FreshTomato) before the upgrade button is pressed. 

Prior to this change, firmware file names are getting truncated on both standard and advanced themes.

### DIFF
--- a/release/src-rt-6.x.4708/router/www/tomato.css
+++ b/release/src-rt-6.x.4708/router/www/tomato.css
@@ -1071,7 +1071,10 @@ img[id^="barnoise_"] {
 	width: 23%;
 }
 
-.upgrade-file,
+.upgrade-file {
+	width: 440px;
+}
+
 .import-file,
 #qos-cl-grid .x2c {
 	width: 41%;


pedro0311:
Could you please use % instead of hard pxs (problems on mobile)?

wellloaded:
Every browser on mobile has a "desktop-mode" perhaps that would do?
I did think of setting a percentage, but for wide screens it really moves the button too far out on the right-hand side and IMO it doesn't look good at all.
Let me have a think on this one, perhaps moving the Upgrade button to the next line is a smart solution overall?

pedro0311:
So maybe max-width (px) + width (%)?

wellloaded:
Possible but we really need to test this properly, I'm worried specifying both *might* cause more issues than results due to the way different browsers interpret this.

We don't have many pages that place content, outside of a table, that far on the right.

Somehow I have the feeling that the “most compatible” way to implement this is to just move the upgrade button to the next line and it can still stay on the right where it is. Let me try...

pedro0311:
OK, sure.

wellloaded:
Creating a new clean one

PR #59, opened 2025-02-13
dnsmasq: fixes in dnsmasq Tomato-helper patch

lancethepants:
Remove trailing carriage return that fgets returns from reading the delete file. Cleans input and fixes issue with musl libc.
Use inet_aton directly instead of inet_addr which is deprecated. inet_addr just calls inet_aton anyway.
Add lease_update_dns(0) to avoid use after free issue which on occasion shows garbled hostname after deleting a lease.


lancethepants:
Patching patches always looks goofy, so verify it all looks good to you.  I did compile and run on R7000.  I tried to format the patch they way you like them.

pedro0311:
Thanks!

PR #60, opened 2025-02-15
tomato.js - allow to control the display of One-Off option in refresh list

wellloaded:
In some cases (e.g. device-list) the one-off option is somehow unwanted as the user would want the refresh to be off or on only.


wellloaded:
I just thought that we need to retain the "full off" option, let me submit a modification

pedro0311:
And please use tabs ;)

PR #62, opened 2025-02-15
tomato.css - extend maximum number of visible characters in upgreade/filename

wellloaded:
Prevents filename truncation for long filename in the upgrade "select file" field
(tested with basic and advanced themes and also on android)


pedro0311:
Can you convert it to tabs, please (for consistency)?
I just tested online editing on github and can confirm that tabs can be used.

PR #63, opened 2025-02-16
Update tomato.css - extend maximum number of visible characters in upgrade/filename

wellloaded:
Resubmitting after the tab correction.


pedro0311:
Thanks!

PR #64, opened 2025-02-16
GUI: admin-upgrade.asp - layout improvement

wellloaded:
- Re-organised layout
- Allows for long filenames to be fully visible now
- moved "clear NVRAM" above the Upgrade button as the option is dependent on it

(apologies about the prev PR, there was indeed an issue with the tabulation within Visual Studio Code as the program by default replaces tabs with spaces). Go figure...


pedro0311:
Thanks!

PR #65, opened 2025-02-17
Network discovery v2.62

wellloaded:
Changelog:
- Code cleaning and optimisation for speed
- Allows to scan LAN as well as WAN interfaces
- Introduces a limit of minimum netmask /22 (e.g. /21 is ignored)
- Adds netcat as a scanning method
- Adds "all" method (round-robin)
- It now excludes existing IPs from the list of scan targets
- Adds Sanitize option to clear up the ip neighbouring table
- Allows to define the aggressiveness of the scan via the Max Probes option (5-200, def 60)
- A change of parameter triggers an automatic  discovery kill + new discovery on new parameters
- Adds control to prevent overlapping discoveries if no parameters have changed
- Support for debug (/tmp/discovery.debug) setting manually "nvram set discovery_debug=1"

Performance:
LAN+WAN in arping + 60 sessions + clear:
~=10 seconds on RT-AC1900P
~=20 seconds on RT-N16

Wiki page of reference:
https://wiki.freshtomato.org/doku.php/status-devices

### DIFF
--- a/release/src-rt-6.x.4708/router/httpd/tomato.c
+++ b/release/src-rt-6.x.4708/router/httpd/tomato.c
@@ -2110,15 +2110,56 @@ static void asp_css(int argc, char **argv)
 }
 
 #if defined(TCONFIG_BCMARM) || defined(TCONFIG_MIPSR2)
-static void asp_discovery(int argc, char **argv)
-{
-	char buf[64] = "/usr/sbin/discovery.sh ";
+static void asp_discovery(int argc, char **argv) {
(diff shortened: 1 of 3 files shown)


pedro0311:
Is it ready now?

wellloaded:
v2.62 is fully operational.
There will be literally 4 lines of script that we are still discussing and are potentially going to result into a little modification; but these are optimisation only. It's safe to import this for the time being. I'll do another small PR with the very final changes later on.

PR #66, opened 2025-02-20
Network discovery v2.64

wellloaded:
Network Discovery v2.64

Changelog:

-    Code cleaning and optimisation for speed
-    Allows to scan LAN as well as WAN interfaces
-    Introduces a limit of minimum netmask /22 (e.g. /21 is ignored)
-    Adds netcat as a scanning method
-    Adds "all" method (round-robin)
-    It now excludes existing IPs from the list of scan targets
-    Adds Sanitize option to clear up the ip neighbouring table
-    Allows to define the aggressiveness of the scan via the Max Probes option (5-200, def 60)
-    A change of parameter triggers an automatic discovery kill + new discovery on new parameters
-    Adds control to prevent overlapping discoveries (if no parameters have changed) via pid management
-    Defines process traps to better integrate with the OS
-    Support for debug (/tmp/discovery.debug) setting manually "nvram set discovery_debug=1"

Wiki page of reference:
https://wiki.freshtomato.org/doku.php/status-devices

### DIFF
--- a/release/src-rt-6.x.4708/router/httpd/tomato.c
+++ b/release/src-rt-6.x.4708/router/httpd/tomato.c
@@ -2110,16 +2110,57 @@ static void asp_css(int argc, char **argv)
 }
 
 #if defined(TCONFIG_BCMARM) || defined(TCONFIG_MIPSR2)
-static void asp_discovery(int argc, char **argv)
-{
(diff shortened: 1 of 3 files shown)


wellloaded:
@pedro0311 please notice the . nvram_ops approach was intentionally avoided as there's just a single nvram get, where this script is meant for speed.

PR #69, opened 2025-02-22
www: adds SSID [ Scan ] button to basic-network.asp and advanced-wlanvifs.asp

wellloaded:
Adds the [ Scan ] Button to SSIDs input field selection.
This applies to both `basic-network.asp` and `advanced-wlanvifs.asp` when wireless Mode is different from Access Point.
 
 

NOTE: This commit also introduces a change to make the repo's version of `basic-network.asp` usable for `mount --bind ` operations.

### DIFF
--- a/release/src-rt-6.x.4708/router/www/advanced-wlanvifs.asp
+++ b/release/src-rt-6.x.4708/router/www/advanced-wlanvifs.asp
@@ -45,9 +45,9 @@ var wl_modes_available = [];
 
 var wmo = {'ap':'Access Point','apwds':'Access Point + WDS','sta':'Wireless Client','wet':'Wireless Ethernet Bridge','wds':'WDS'
 /* BCMWL6-BEGIN */
-	   ,'psta':'Media Bridge'
+		,'psta':'Media Bridge'
 /* BCMWL6-END */
-	   };
+		};
 var macmode = {'disabled':'Disabled','deny':'Block','allow':'Permit'};
 
 var tabs = [['overview','Overview']];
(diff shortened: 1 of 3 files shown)


HommeOursPorc:
If this is the same version from the one of the thread, there's several issues with it. 

1. Wireless Ethernet Bridge mode is greyed out and cannot be selected on any radio anymore. When selecting an other mode, it is impossible to go back to Wireless Ethernet Bridge.

2. Scan button doesn't appear for the second radio.

3. When selecting WEB or MB, then going back to AP mode, the scan button and dropdown menu doesn't disappear, so the scan button and dropdown menu can still be selected in AP mode.

wellloaded:
> If this is the same version from the one of the thread, there's several issues with it.
> 
>     1. Wireless Ethernet Bridge mode is greyed out and cannot be selected on any radio anymore. When selecting an other mode, it is impossible to go back to Wireless Ethernet Bridge.
> 
>     2. Scan button doesn't appear for the second radio.
> 
>     3. When selecting WEB or MB, then going back to AP mode, the scan button and dropdown menu doesn't disappear, so the scan button and dropdown menu can still be selected in AP mode.

1) Thanks, but are you 100% certain this is a "new behaviour" introduced by this change? I experience a similar issue when in wireless client, as I need to define this in the WAN settings, if I change in the wireless interface to something else I can't go back to wireless client. e.g. can you unmount the files and see if this is still happening?

2) what is the Wireless mode of the second radio?

3) This might be a little issue and will look into it. As mentioned in the forum I need to change the code to use the "case" condition instead of introducing an ad hoc control for the Scan button.

Thanks

HommeOursPorc:
> > If this is the same version from the one of the thread, there's several issues with it.
> > ```
> > 1. Wireless Ethernet Bridge mode is greyed out and cannot be selected on any radio anymore. When selecting an other mode, it is impossible to go back to Wireless Ethernet Bridge.
> > 
> > 2. Scan button doesn't appear for the second radio.
> > 
> > 3. When selecting WEB or MB, then going back to AP mode, the scan button and dropdown menu doesn't disappear, so the scan button and dropdown menu can still be selected in AP mode.
> > ```
> 
>     1. Thanks, but are you 100% certain this is a "new behaviour" introduced by this change? I experience a similar issue when in wireless client, as I need to define this in the WAN settings, if I change in the wireless interface to something else I can't go back to wireless client. e.g. can you unmount the files and see if this is still happening?

Yes, just tested it, on the original page the Wireless Ethernet Bridge option is always available no matter what I tried. In your latest Pastebin version Wireless Ethernet Bridge is always greyed out and unavailable (I didn't had time to test the version uploaded on the forum yet). 

>     2. what is the Wireless mode of the second radio?
Access Point

>     3. This might be a little issue and will look into it. As mentioned in the forum I need to change the code to use the "case" condition instead of introducing an ad hoc control for the Scan button.
> 
> 
> Thanks

You're welcome 😁

PR #70, opened 2025-02-22
www: status-devices.asp - added Notes section

wellloaded:
Added Notes section including elements from the discovery v2.64.


pedro0311:
Added as a separate commit because of needed changes: https://github.com/FreshTomato-Project/freshtomato-arm/commit/a46130f85fc24722f38a04b16c4b0ddb0651c77a
Thanks!

PR #74, opened 2025-03-04
libshared: Default DSCP fix to off.

lancethepants:
Looks like comacast fixed this issue back in 2023.
https://forum.openwrt.org/t/cake-w-dscps-cake-qos-simple/147087/252


pedro0311:
Thanks!

PR #78, opened 2025-03-19
httpd: iptraffic.c: printf for unsgined int should use %u

lancethepants:


lancethepants:
Tested on R7000

pedro0311:
Thanks! Interesting that gcc didn't complain about it...

lancethepants:
> Thanks! Interesting that gcc didn't complain about it...

Right!

PR #82, opened 2025-03-30
Updated wireless Icons for status-devices.asp

wellloaded:
Some minors restyle.
Dropping .4 and .0 from the icons and adding GHz instead.
2.4 -> 2GHz
5.0 -> 5GHz


wellloaded:
I'll push a new PR that uses SVG icons instead.

PR #84, opened 2025-03-31
GUI: advanced-vlan.asp - Header correction

wellloaded:
A bridge maps to a VLAN in this case and not to an Ethernet port. So the “VLAN to bridge mapping “is the correct terminology.


pedro0311:
Thanks!

PR #87, opened 2025-04-14
GUI: basic-time.asp - NTP Client added 4th placeholder field for Custom upstream server.

xixix-ft:
The expectation is to pull time from a NTP server/IP within the LAN if WAN is not available

Looks like the .asp is the only place that needs adjustment as I'm able to:
```
$ nvram set ntp_server='0.us.pool.ntp.org 1.us.pool.ntp.org 2.us.pool.ntp.org 192.168.0.14'
$ nvram commit
$ service ntpd restart
```
and properly setups **ntp.conf**:
```
$ cat /etc/ntp.conf

server 0.us.pool.ntp.org
server 1.us.pool.ntp.org
server 2.us.pool.ntp.org
server 192.168.0.14
```


pedro0311:
Thanks!

PR #89, opened 2025-04-17
www: status-devices.asp: Add option to hide WAN devices

lancethepants:
@wellloaded thought this would be a nice feature to upstream.


pedro0311:
Thanks!

PR #91, opened 2025-04-26
- httpd: misc.c: Issue with CPU Usage NaN% in Status Overview page

xixix-ft:
Few days after FT router reboots, everything seems to work fine with the CPU Usage % metrics in Status Overview page.
After weeks at router full workload capacity, the percentage turns into a NaN%... "inconsistently"⚠️

https://www.linksysinfo.org/index.php?threads/issue-with-cpu-usage-nan-in-status-overview-page-a-walkthrough-arm-asus-rt-ac3200.79175/post-358580


pedro0311:
Thanks!

PR #94, opened 2025-05-08
www: bwlimit.asp: iterate over MAX_BRIDGE_ID to create bridge classes

lancethepants:


pedro0311:
Well, I thought about it, but never had time to do it. Thanks!

PR #106, opened 2025-06-09
GUI: status-devices - correction to the 2.4 and 5.x SVG icons

wellloaded:
- Changed font to fix compatibility issue
- "2" text icon renamed to "2.4"

Thanks to @WildFireSG for the suggestion.


lancethepants:
Something I've noticed on the status devices page is there is a "flicker" I will call it.  You can see a kind of ghost of a page layout like the page is completely different for a second, and then it reshapes or corrects itself at the end of load.  I haven't investigated it yet but I'm wondering if it has something to do with the change to svg icons.

wellloaded:
> Something I've noticed on the status devices page is there is a "flicker" I will call it. You can see a kind of ghost of a page layout like the page is completely different for a second, and then it reshapes or corrects itself at the end of load. I haven't investigated it yet but I'm wondering if it has something to do with the change to svg icons.

I've never noticed this on my side. I'm wondering if it has anything to do with the cache? What browser are you using?

BTW, without me creating a PR for T64 the new 6GHz icon is this one (to retain font/style consistent).
wl60

lancethepants:
I'm not seeing it in Firefox, but I am seeing it in chrome.

wellloaded:
> I'm not seeing it in Firefox, but I am seeing it in chrome.

I could reproduce this in Edge. According to AI this is a well-known behaviour in Edge due to the way it handles the DOM. Some workarounds given are "colourful" but ultimately it's suggesting to avoid reloading the full page content. I'm not sure what we could do, really, as it seems to suggest "Edge it's Edge". Do you think this was introduced by the svg icons only? E.g. reverting the svg icon commit this doesn't happen?

lancethepants:
It seemed like it happened around that time but I'm not sure yet.  I'll test it out with the old icons when I get a sec.

pedro0311:
I see no problems with old icons on FF, on Chrome icons are redrawn with every automatic page refresh (so they disappear for a split second - it doesn't look good), but in both cases (FF, Chrome) 2/5 icons are bigger than old ones (WAN/LAN/etc.).
Clipboard01


I'll check out the new ones.
//EDIT: the new ones are also bigger (this time ss is from advanced theme):

Clipboard02

wellloaded:
> I see no problems with old icons on FF, on Chrome icons are redrawn with every automatic page refresh (so they disappear for a split second - it doesn't look good), but in both cases (FF, Chrome) 2/5 icons are bigger than old ones (WAN/LAN/etc.). Clipboard01
> 
> I'll check out the new ones. //EDIT: the new ones are also bigger (this time ss is from advanced theme):
> 
> Clipboard02

The size increase from 25px to 28px was intended as they add some healthy padding to each line (IMO at least). If you still see old icons... try to refresh the cache? The svg, AFAIR were always produced at 28px.

image


I doubt the size to be an issue regardless though, the only thing I can think about is for svg to have a different cache interaction compared to gif. Out of my tests, the issue is indeed visible on Edge and Chrome, where FF behaves as expected.

pedro0311:
OK, found it: https://github.com/FreshTomato-Project/freshtomato-arm/commit/9b7f810f8687a5674db447c86448252a68c28004

Another question: can you remove not needed spaces and LFs from that icons?
Or maybe you want me to do it?

pedro0311:
BTW, talking about optimization: svg icons are not good in it, because our httpd regarding other svg files needed for BW and IPT uses 0 cache for them. So they are always fetched.

wellloaded:
> BTW, talking about optimization: svg icons are not good in it, because our httpd regarding other svg files needed for BW and IPT uses 0 cache for them. So they are always fetched.

I was incidentally looking into this and came to the same conclusion. However I'm sure there's a way to en able/disable svg caching on a page basics. Let me look for a solution.

Could you point me out where precisely we have an explicit exclusion of svg caching at code level?

pedro0311:
Added as a separate commit, thanks!
https://github.com/FreshTomato-Project/freshtomato-arm/commit/a4c9a4106d6ce152ef2feaf72a95ed7a1ff8318e

wellloaded:
Wait, did this fix the flickering?
I'm reading a possible workaround is to embed the svg in the .asp this way they are treated as text and not external files.

PR #107, opened 2025-06-23
www: fix qr-code svg

lancethepants:


lancethepants:
It appears to me that the qr code svg doesn't appear correctly, at least in chrome. Maybe you'll find a better solution, but this is what I've thrown in for the time being.

pedro0311:
I really don't know, why you guys have problem with that icon...
I checked on FF LTS, latest FF, old Chrome (win7), latest Chrome, Opera, latest Pale Moon and on each of these browsers the QR icon looks fine.
Anyway, thanks!

Issue #88, opened 2025-04-17
Suggested text edits to VLAN menu Notes section


whateverusernameisnottaken:
I suggest the following edits to the VLAN menu's Notes section:

* VLAN - the VLAN's unique identifier. Valid settings are 0 to 15. This is locally significant and should be considered as a GUI reference only.
* VID - The label that identifies to which VLAN the frame belongs. Valid entries are 1 to 4094. When set, binds VLAN & VID (Such as: VLAN: 4 to VID: 4) on a port basis.
The VLAN and VID number of reference can match but needn't. The special VID "0" means no label and therefore no VLAN configuration applied to frames.
* Ports 1 to 4 and WAN - specify which Ethernet ports will be members of this VLAN.
* Tagged - enables 802.1Q tagging of Ethernet frames on a particular port. This allows multiple VLANs to be configured one given port.
* Default - the VLAN ID assigned to untagged frames received by the router.
* Ethernet-to-bridge mapping - determines the binding between VLAN and physical interfaces such as WANx or LANx. A mapping with "none" is considered unused.

This sentence doesn't make sense as is:
Bridge $wireless_if to $lan_bridge - For each wireless interface define (physical or virtual) specify to what LAN bridge this has to map

Did you mean: "Bridge $wireless_if to $lan_bridge - For each (physical or virtual) wireless interface, specify to what LAN bridge it will be mapped"

Other relevant notes/hints:
* It's wise to cross-check settings in the Basic/Network and Advanced/Virtual Wireless menus.
* Be mindful that some FreshTomato releases might use VID: 0 as default others VID: 1.
* One VID must be assigned to the WAN bridge.
* One VID must be selected as the default VLAN.
* To prevent 802.1Q compatibility issues, avoid using VID: "0". 802.1Q specifies that frames with a "0" tag don't belong to any VLAN. The tag contains only user priority information.
* If you trunk with other vendors, note that some discourage from using VID: "1" as it may be specially reserved (for management purposes).

pedro0311:
It would be nice, if @lancethepants and/or @HommeOursPorc could review these changes.

HommeOursPorc:
I'll look it up soon. 

First thing I noticed:

"Be mindful that some FreshTomato releases might use VID: 0 as default others** VID: 1 for **LAN**."

Not sure how to write it though 🤔

Maybe "as default for LAN, others VID: 1"

[...]

PR #109, opened 2025-06-25
rc: firewall.c: prevent devices from accessing router services of oth…

lancethepants:
…er bridges/lans

### DIFF
--- a/release/src-rt-6.x.4708/router/rc/firewall.c
+++ b/release/src-rt-6.x.4708/router/rc/firewall.c
@@ -1124,6 +1124,10 @@ static void filter_input(void)
 	char *hit;
 	int i, n;
 	char *p, *c;
+	char lanN_ifname[] = "lanXX_ifname";
+	char lanN_ifname2[] = "lanXX_ifname";
+	char lanN_ipaddr2[] = "lanXX_ipaddr";
+	char br, br2;
 
 #ifdef TCONFIG_BCMARM
 	/* 3 for filter */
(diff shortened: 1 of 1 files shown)


lancethepants:
This is request for Comments

I've received several complaints about devices on one network, let's say with IP Address 192.168.2.10, being able to access the WebUi (among other things) on the router's other networks lan IPs, say 192.168.1.1.  This extends to all services the router provides.  You may for example tell samba to only listen on br0, but devices on br1 can access the samba share through the router's IP address on br0, even though on the file sharing page br1 has not been selected for file sharing.

The following rules on the filter input chain unconditionally allow all traffic destined for the router regardless of which network it's coming from
-A INPUT -i br0 -j ACCEPT
-A INPUT -i br1 -j ACCEPT
-A INPUT -i br2 -j ACCEPT
-A INPUT -i br3 -j ACCEPT

This commit then adds rules like the following (depending on your configuration) immediately before the ACCEPT rules  from above, to block access to all router lan IP addresses except the netowrk that a device belong to.

-A INPUT -i br0 -d 192.168.2.1 -j DROP
-A INPUT -i br0 -d 192.168.3.1 -j DROP
-A INPUT -i br0 -d 192.168.4.1 -j DROP
-A INPUT -i br1 -d 192.168.1.1 -j DROP
-A INPUT -i br1 -d 192.168.3.1 -j DROP
-A INPUT -i br1 -d 192.168.4.1 -j DROP
-A INPUT -i br2 -d 192.168.1.1 -j DROP
-A INPUT -i br2 -d 192.168.2.1 -j DROP
-A INPUT -i br2 -d 192.168.4.1 -j DROP
-A INPUT -i br3 -d 192.168.1.1 -j DROP
-A INPUT -i br3 -d 192.168.2.1 -j DROP
-A INPUT -i br3 -d 192.168.3.1 -j DROP


This is similar to the rules on the FORWARD chain that block communication between devices on different networks.

-A FORWARD -i br0 -o br1 -j DROP
-A FORWARD -i br0 -o br2 -j DROP
-A FORWARD -i br0 -o br3 -j DROP
-A FORWARD -i br1 -o br0 -j DROP
-A FORWARD -i br1 -o br2 -j DROP
-A FORWARD -i br1 -o br3 -j DROP
-A FORWARD -i br2 -o br0 -j DROP
-A FORWARD -i br2 -o br1 -j DROP
-A FORWARD -i br2 -o br3 -j DROP
-A FORWARD -i br3 -o br0 -j DROP
-A FORWARD -i br3 -o br1 -j DROP
-A FORWARD -i br3 -o br2 -j DROP

So far I haven't seen any downsides to including these new set of rules, but I feel like it's a big enough change that I want more eyeballs looking at it. Do you see any issues with it?  I don't imagine it would get in the way of an VPN routing or setups, but I haven't tested all scenarios.

lancethepants:
Or maybe there's a better way to write the rules and accomplish the same thing?

MarsTomato:
hi lance,
see also IPv6
https://github.com/FreshTomato-Project/freshtomato-arm/commit/76e33d86a3764b807e8ff83779c6f997d63a9af3

mmhh... there was a discussion about it on the forum and also the issue tracker at the bitbucket arm repo (deleted) from pedro, maybe i can find it later.

Yes, FORWARD we allow only br0 to br0 by default for example AND 
INPUT is allowed anything (IPv4 & IPv6)

I think it was the simplest default setup for different use cases, but i do not remember everything right now :-) --> too hot

pedro0311:
I think if we are going to make such a change to FW, there should be an option to enable/disable this feature in the GUI, but I don't know if it should be enabled or disabled by default.
And what about @MarsTomato  commit regarding IPv6 then?

scoobasteve83:
https://www.google.com/search?q=should+bridges+on+a+router+have+the+same+mac+id

Solved my issue with having to connect many times over to get internet connection by changing this value. Just an observation connectivity is better as well so IDK. Try it before you knock it.

lancethepants:
> I think if we are going to make such a change to FW, there should be an option to enable/disable this feature in the GUI, but I don't know if it should be enabled or disabled by default. And what about @MarsTomato commit regarding IPv6 then?

I think that is reasonable.  Disabled by default I think is fine just to approach things cautiously, with the option put under the advanced-firewall page.  Probably IPv6 needs the same treatment and enablement with the same setting.

MarsTomato:
> Disabled by default I think is fine just to approach things cautiously, with the option put under the advanced-firewall page

yes, think so too 👍 

> Probably IPv6 needs the same treatment and enablement with the same setting.

yes

pedro0311:
@lancethepants : would you be so kind and update your commit to our needs?
Just add nvram variable "fw_something" or whatever, add it do tomato.c, defaults.c, and advanced-firewall.asp + your changes in firewall.c + switch in @MarsTomato commit https://github.com/FreshTomato-Project/freshtomato-arm/commit/76e33d86a3764b807e8ff83779c6f997d63a9af3 .
Thanks!

pedro0311:
@lancethepants : also if you could review these changes to VLAN page: https://github.com/FreshTomato-Project/freshtomato-arm/issues/88 I would be grateful.

PR #130, opened 2025-08-28
Update CN tag to reflect SAN instead

scoobasteve83:
The problem is the title says CN

You are changing the SAN 

So if you look at Cert CN don't get changed it is SAN

which is proper.


scoobasteve83:
ermmm I guess there is some odd merge

PR #132, opened 2025-08-30
Update status-devices.asp HTTP & HTTPS Links

scoobasteve83:
Please double check the code but it may be right or wrong...

I hope this helps RS232's request.

to show http and https links

IDK really if it is correct code might be a little off and need editing but this is just a thought on how to accomplish it.

Not a Developer by trade have no expirence other than when I need something I try to make it if it don't exist but what the heck I see folks asking for help so might as well try and do what I can learn as I go ;)

### DIFF
--- a/release/src-rt-6.x.4708/router/www/status-devices.asp
+++ b/release/src-rt-6.x.4708/router/www/status-devices.asp
@@ -91,7 +91,10 @@ dg.populate = function() {
 	wl_info = [];
 
 	for (i = 0; i < list.length; ++i) {
-		list[i].ip = '';
+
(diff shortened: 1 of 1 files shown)


scoobasteve83:
my thinking is we can show the IP then right below it have http | https as clickable links

so it would look like this:
192.168.1.44
http | https  <these would be clickable links

Like I say the code could be wrong I don't know how to quickly test and verify either where the code is put or maybe the populate function only populates the number of columns and this will have to be hooked in somewhere else IDK its kinda confusing like there is code missing in some spots maybe its in tomato.js or tomato.css

PR #137, opened 2025-09-03
For rs232 request http and https links.

scoobasteve83:
Updated code to make safer and fix syntax and open in new tab as requested by @pedro0311 

This should allow the clickable links I believe have them open in new tabs and be safer.

Would be nice if someone could test a build with this I did ask lance. Don't have a rig to test got some parts but need a few more likely won't be till after xmas I'm guessing but I'm working toward getting there.


scoobasteve83:
RS232 is going to make his own patch!

PR #142, opened 2025-09-07
Update targets.mak: dir868lwg: remove:

txnative777:
- this extra target was left out as it still had "NO_SSH=y".


pedro0311:
Thanks, I see a glitch here in github, something's wrong here - it shouldn't be there...

PR #144, opened 2025-09-12
tomato.css - improved to print and printscreen in dark-mode

wellloaded:
This change improves the quality of printscreen and physical prints while the GUI is in .alt-mode (a.k.a. dark mode) and the brightness/sepia filter is applied.


pedro0311:
Thanks!

PR #158, opened 2025-10-27
GUI: status-overview.asp - Only displaying unsecured WiFi warning in AP mode

wellloaded:
As reported here:

https://www.linksysinfo.org/index.php?threads/disable-the-overview-wifi-security-disabled-warning-when-the-relevant-interface-is-in-wireless-client-mode.79444/


pedro0311:
Thanks!

PR #162, opened 2025-11-03
shared: misc.c: increase ifnames buffer size

lancethepants:
Avoids buffer overflow when using > 4 networks


pedro0311:
Right, thanks!

PR #166, opened 2025-11-10
others: mwwatchdog: fix operator precedence bug that could add cron j…

lancethepants:
…ob when mwan_cktime=0


lancethepants:
Removing the braces changes the logic. switch4g unconditionally calls mmwwatchdog which can enable adding cronjob even if mwan_cktime=0

pedro0311:
Thanks!

PR #171, opened 2025-12-07
Add Bridge Gateway Isolation + UI (IPv4 only atm), IPv6 bridge isolation, and IPv6-aware advanced-access.asp

lancethepants:
Bridge Gateway Isolation blocks devices from accessing gateway IPs on other bridges. This prevents devices gaining access to services only enabled for other bridges/lans. (ie samba, media server etc)
Currently IPv4 only as it relies on network settings for all bridges to be stored in nvram.

IPv6 bridge isolation prevents cross bridge IPv6 communication between clients like has existing for a long time with IPv4.


pedro0311:
Thanks!

MarsTomato:
good work :-)

PR #173, opened 2025-12-12
Improved IPv6 support

lancethepants:
Added support for configuring ia-na and ia-pd ID
Iterate over MAX_BRIDGE_ID for interfaces


pedro0311:
Nice one, thanks @lancethepants !

PR #175, opened 2025-12-14
GUI: advanced-vlan.asp overhaul and new ethernet.svg

wellloaded:
- It is now possible to **mount --bind** the page
- Removed "tag N" column and unified the settings under the port column
- Layout improvements
- Simplified Notes
- Removed all the eth_*.gif icons
- status-overview.asp adjustments for ethernet.svg
- Added a single ethernet.svg image supporting the following parameters:
```
speed=
    0 > off (default)
    10 > amber
    100 > green
    1000 > blue
    2500 > red
    5000 > purple
    10000 > white
duplex=
    HD > light grey body (default)
    Half > same as above but thought for T64
    FD > dark grey body
    Full > same as above but thought for T64
flash=
    in fraction of a second, e.g. "0.7" > overrides the hardcoded timing for flashing
caption=
    "text" > text within the icon
```


MarsTomato:
I like the two column approach 🫣

wellloaded:
> I like the two column approach 🫣

I respect that.

The angle where I came from was to think of devices (including T64) that have 8 ports and potentially more, e.g. the Asus RT-88U has 9 ports that means 18 columns. + VLAN + VID + default + Bridge. That's a lot of columns to squeeze into a single tomato frame.

Another "out of the box" idea would be to rotate the table vertically so that it scales automatically in page length based on the number of ports. That's more work, but everything is possible.

scoobasteve83:
In truth I didn't like it until I tried it.... Once you bind the page and look at it and mess with it you kinda understand its usefulnees the only thing I didn't like was the lables.... On and TAG should've been tagged and untagged to stick with what is VLAN standard. 

Finally the first column label should be FWID cause it is really only an id for the firmware it has nothing to do with VLAN that part was confusing to me setting up VLAN's.

Good work @wellloaded your effortt was appreciated! 

Would be nice to add a column for VLAN labels and shorten the number of colums or maybe make the firmware id user enterable labels instead. So you can name them IOT, TVs, Lightswitches, Cameras, Kids, Phones etc... more lan bridges would be nice to make use of the 15 VLAN groups so possibly cutting the number of vlans to 8 and allowing 8 bridges would take up the same amount of NVRAM IDK. I realize space is scarce. 

Been kicking some ideas around looking for feedback.

HommeOursPorc:
Putting this here for awareness. Not long ago I came up with an idea to add a network summary to the VLAN page to make configuring easier. Is this something that could be considered for the VLAN page overhaul? 

I already provided working code in the thread on the forum. It is AI generated though.

https://www.linksysinfo.org/index.php?threads/advanced-vlan-connections-infos.79085/

PR #189, opened 2026-01-03
nvram: new fname variable to store FW filename during the flashing process

wellloaded:
Using the identical logic used by the build.mak this new init_firmware_filename() function identifies the output filename as produced by the build.mak recipe and stores it into a new **fname** nvram variable.

```
root@sparrow:/# nvram get fname
freshtomato-RT-AC1900P-K26ARM-2025.5-AIO-64K.trx

```

This variable is forcedly overwritten at each flashing (only) with the idea to always have the current firmware filename made available to the system.

ne6nSluZr3

### DIFF
--- a/release/src-rt-6.x.4708/router/rc/init.c
+++ b/release/src-rt-6.x.4708/router/rc/init.c
@@ -120,6 +120,7 @@ static void restore_defaults(void)
 
 	nvram_set("os_version", tomato_version);
 	nvram_set("os_date", tomato_buildtime);
+	init_firmware_filename();
 
 #if defined(TCONFIG_BLINK) || defined(TCONFIG_BCMARM) /* RT-N+ */
 	/* Adjust et and wl thresh value after reset (for wifi-driver and et_linux.c) */
@@ -857,6 +858,40 @@ static int wlshutdown_ethx_rtac5300(void)
 	return dirty;
 }
 #endif /* TCONFIG_AC5300 */
+
+static void init_firmware_filename(void)
+{
+	char *model, *os_ver;
+	char ver[16], branch[16], build[64];
+	char fname[256];
+
+	/* Get model from odmpid (clean model name without manufacturer prefix) */
+	model = nvram_safe_get("odmpid");
+	if (!*model)
+		model = nvram_safe_get("productid");
+	if (!*model)
+		model = nvram_safe_get("model");
+
+	/* Parse os_version: "2025.5 K26ARM USB AIO-64K" */
+	os_ver = nvram_safe_get("os_version");
+	if (sscanf(os_ver, "%15s %15s %*s %63s", ver, branch, build) == 3) {
+		/* Construct filename matching build.mak pattern:
+		 * freshtomato-MODEL-BRANCH-VERSION-BUILD.trx */
+		snprintf(fname, sizeof(fname), "freshtomato-%s-%s-%s-%s.trx",
+			model, branch, ver, build);
+	}
+	else {
+		/* Fallback if os_version is malformed (should never happen in normal builds) */
+		snprintf(fname, sizeof(fname), "freshtomato-%s-unknown-unknown-unknown.trx", model);
+		syslog(LOG_WARNING, "Failed to parse os_version: %s", os_ver);
+	}
+
+	/* Only update if filename changed */
+	if (!nvram_match("fname", fname)) {
+		nvram_set("fname", fname);
+		syslog(LOG_INFO, "Firmware filename: %s", fname);
+	}
+}
 
 static void init_lan_hwaddr(void)
 {
--- a/release/src-rt-6.x.4708/router/shared/defaults.c
+++ b/release/src-rt-6.x.4708/router/shared/defaults.c
@@ -338,6 +338,7 @@ struct nvram_tuple bsd_defaults[] = {
 
 struct nvram_tuple router_defaults[] = {
 	{ "restore_defaults",		"0"				, 0 },	/* Set to 0 to not restore defaults on boot */
+	{ "fname",				""				, 0 },	/* Populate filename used during flashing */
 
 	/* LAN H/W parameters */
 	{ "lan_hwnames",		""				, 0 },	/* LAN driver names (e.g. et0) */
--- a/release/src-rt-6.x.4708/router/www/admin-upgrade.asp
+++ b/release/src-rt-6.x.4708/router/www/admin-upgrade.asp
@@ -18,7 +18,7 @@
 
 <script>
 
-//	<% nvram("jffs2_on,jffs2_auto_unmount,remote_upgrade"); %>
+//	<% nvram("jffs2_on,jffs2_auto_unmount,remote_upgrade,fname"); %>
 
 //	<% sysinfo(); %>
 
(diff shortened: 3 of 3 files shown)


wellloaded:
Fixed thanks, at every boot there's now a control on the fname nvram content against the built-in filename. So only if different the fname in nvram gets overwritten preventing unnecessary nvram writes.

```
if (!nvram_match("fname", fname)) {
...
```

HommeOursPorc:
If the erase nvram checkbox is checked, after the reboot the fname variable will be empty. What will be displayed then?

wellloaded:
> If the erase nvram checkbox is checked, after the reboot the fname variable will be empty. What will be displayed then?


  * The filename is compiled into the firmware.
  * If the `fname` variable is empty or differs from the built-in filename:

    * At boot (init stage), the system checks the `fname` content.
    * The system determines whether to:

      * Write or overwrite the filename.
      * Ignore it if it's already correct.

* **Result**: As long as the system is running, the filename remains available in the `fname` variable.

scoobasteve83:
hrmmm.... Should be done with .cfg as well putting a fw build spec in each cfg. This way you could have a script that would be able to properly sanitize them like a ffg handler/cleanup script to any version of firmware without issue but this could be a simple marker that is placed in all .cfg backups. This way you can restore cfg to any version and a script would take over and make any corrections to the cfg to make it restorable on the newer version. The script should get updated and cleared every year and only retain the last update of the previous calender year on what to do with .cfg than that way if someone is really behind all they need to do is a daisy chain upgrade with the last build of each year they would be behind. So basically flash 2024.6 load cfg then save new cfg then flash 2025.5 save new cfg then load 2026.x and load cfg and it would work. This way the script don't grow to be a monstrosity that is impossible to maintain.

HommeOursPorc:
It is a nvram variable, so it will be saved in the cfg file with all the other nvram variable. cfg is really only a `nvram show` command with some extra steps to "encrypt" the content. It is not really encrypted but by lack of other words, I'll call it that.

HommeOursPorc:
FreshTomato version is already included in every cfg backup file btw. It is in the os_version variable IIRC.

scoobasteve83:
encoded same pile well sheesh if version is in cfg nothing is really stopping a handler from being created to restore a backup properly between versions technically speaking cool deal.

If cfg is this version change these nvram key names or remove these nvram keys and values before applying the configuration interesting thanks homme!

pedro0311:
On my Asus RT-AC3200 (filename used: "freshtomato-RT-AC3200-K26ARM7-2025.5.001-beta-AIO-128K.trx") it doesn't work:

>root@router:/tmp/home/root# nvram get odmpid
>ASUS
>root@router:/tmp/home/root# nvram get productid
>root@router:/tmp/home/root# nvram get model
>RT-AC3200
>root@router:/tmp/home/root# nvram get os_version
>2025.5.001 K26ARM7 beta USB AIO-128K

And I get: 

`freshtomato-ASUS-K26ARM7-2025.5.001-USB.trx`

So the logic here must be changed (now, it will never use "RT-AC3200" string).

pedro0311:
Of course it works with "normal" version (with no "beta" in it, except for "ASUS" instead of "Asus") but we should cover also that case IMO.

HommeOursPorc:
FWIW, here's the "algorithm" I came up with for my router identifying part in my ftup script... Just to give an idea... If this can help somehow... Don't solely rely on it though 😆

```
...
OSVER="$(NG os_version)"
CURVER="$(echo "$OSVER" | awk '{print $1}')"
FULLMODEL="$(NG t_model_name)"
...
echo "$FULLMODEL" | grep -qi "Cisco Linksys" && MAKE="Cisco Linksys" || MAKE="$(echo "$FULLMODEL" | awk '{print $1}')"
TMODEL="$(echo "$FULLMODEL" | awk '{if(NF>=3){{print $(NF-1)"_"$NF}}else{{print $NF}}}' | tr -d "/")"
FIX="$(NG t_fix1)"
ID="$(NG odmpid)"
[ "$FIX" == "$TMODEL" ] && MODEL="$FIX" || { [ "$ID" != "" ] && MODEL="$ID" || MODEL="$FIX" ; }
[ "$MATCH" != "" ] && { [ "$MATCH" == "$MODEL" ] && ACCURACY=100 || MODEL="$MATCH" ; }
SDK="$(echo "$OSVER" | awk '{ for(i=1; i<=NF; i++) { if(match($i, /^K+/)) { print $i; break; } } }')" # eg K26ARM
[ "$CPU" == "mips" ] && {
    MIPSREV="$(echo "$SDK" | cut -d "_" -f1 | sed "s/K26//i")"
    SDK="$(echo "$SDK" | cut -d "_" -f2)"
    NVRAM="$(nvram show | grep "entries,.*bytes" | awk '{print ($3+$6)/1024}')K"
} || {
    NVRAM="$(nvram show 2>&1 >/dev/null | tr -d "()" | awk '{print ($2+$4)/1024}')K"
}
case "$SDK" in
    RTAC) SDKFULL="K26RT-AC";;
    RTN ) SDKFULL="K26RT-N";;
    RT  ) SDKFULL="K26";;
    *   ) SDKFULL="$SDK";;
esac
VAR="$(echo "$OSVER" | awk '{print $NF}')" # eg VPN-64K
}

id_show() {
id
echo "
$FULLMODEL
FreshTomato $OSVER

       MAKE = $MAKE
      MODEL = $MODEL
        CPU = $([ "$MIPSREV" != "" ] && echo "$MIPSREV" || echo "ARM")
        SDK = $SDKFULL    $([ $USB -eq 1 ] && printf "\n        USB = YES")
      NVRAM = $NVRAM
      BUILD = $VAR
    VERSION = $CURVER
...
```

wellloaded:
> Of course it works with "normal" version (with no "beta" in it, except for "ASUS" instead of "Asus") but we should cover also that case IMO.

Would you mind sharing what recipe you are using for testing specifically, so that I can look into it?

pedro0311:
It's easy: https://www.onlinegdb.com/online_c_compiler
with:
```C
#include <stdio.h>

static void init_firmware_filename(void)
{
	char *model, *os_ver;
	char ver[16], branch[16], build[64];
	char fname[256];

	/* Get model from odmpid (clean model name without manufacturer prefix) */
	//model = nvram_safe_get("odmpid");
	model = "ASUS";
	//if (!*model)
	//	model = nvram_safe_get("productid"); // none
	if (!*model)
		//model = nvram_safe_get("model");
		model = "RT-AC3200";

	/* Parse os_version: "2025.5 K26ARM USB AIO-64K" */
	//os_ver = nvram_safe_get("os_version");
	os_ver = "2025.5.001 K26ARM7 beta USB AIO-128K"; //bad
	//os_ver = "2025.5 K26ARM7 USB AIO-128K"; // good
	if (sscanf(os_ver, "%15s %15s %*s %63s", ver, branch, build) == 3) {
		/* Construct filename matching build.mak pattern:
		 * freshtomato-MODEL-BRANCH-VERSION-BUILD.trx */
		snprintf(fname, sizeof(fname), "freshtomato-%s-%s-%s-%s.trx", model, branch, ver, build);
	}
	else {
		/* Fallback if os_version is malformed (should never happen in normal builds) */
		snprintf(fname, sizeof(fname), "freshtomato-%s-unknown-unknown-unknown.trx", model);
		printf("Failed to parse os_version: %s", os_ver);
	}

	/* Only update if filename changed */
//	if (!nvram_match("fname", fname)) {
//		nvram_set("fname", fname);
		printf("Firmware filename: %s", fname);
//	}
}

int main()
{
	init_firmware_filename();

	return 0;
}

```

pedro0311:
I don't know if it's worth the extra work, because there may be more versions of the filename, ie.:

`freshtomato-RT-N18U-K26ARM-2025.4.093-beta-wireguard-dropbear-AIO-64K-NOSMP.trx`

So it's better to leave it as it is now.

pedro0311:
Oh, it's also only for ARM routers? Or both?

pedro0311:
Added as a separate commit because of some cosmetic: https://github.com/FreshTomato-Project/freshtomato-arm/commit/f976a3980ac44fcbfe9d659a24a5ad153ae47018

pedro0311:
Fixed version: https://github.com/FreshTomato-Project/freshtomato-arm/commit/91a33ea311d8b26eab9f4f05955e429eb175c0ba

PR #190, opened 2026-01-04
GUI: VLAN page overhauled and new ethernet.svg icon

wellloaded:
It is now possible to mount --bind the page
Removed tag column and unified the settings under the port column
Layout improvements
Simplified Notes
Removed all the eth_*.gif icons
Added single ethernet.svg supporting:

```
speed=
    0 > off (default)
    10 > amber
    100 > green
    1000 > blue
    2500 > red
    5000 > purple
    10000 > white
duplex=
    HD > light grey body (default)
    Half > same as above but thought for T64
    FD > dark grey body
    Full > same as above but thought for T64
flash=
    in fraction of a second, e.g. "0.7" > overrides the hardcoded timing for flashing
caption=
    "text" > text within the icon
```

status-overview.asp adjustments for ethernet.svg


pedro0311:
We need to decide it, if we're using here 'on compilation' markers like:

```
EXT-SW-BEGIN
EXT-SW-END
EXTSW-NO-BEGIN
EXTSW-NO-END
```

or using javascript variable calculated from nvram values to check for LAN number.
Personally, I prefer the first version (as before) because this way we won't add unnecessary code that only applies to the Asus RT-AC88U.

pedro0311:
What I meant, now we have mix of both - you remove some of `EXT-SW-...` but not all of them.
So let's use them as before.

wellloaded:
> We need to decide it, if we're using here 'on compilation' markers like:
> 
> ```
> EXT-SW-BEGIN
> EXT-SW-END
> EXTSW-NO-BEGIN
> EXTSW-NO-END
> ```
> 
> or using javascript variable calculated from nvram values to check for LAN number. Personally, I prefer the first version (as before) because this way we won't add unnecessary code that only applies to the Asus RT-AC88U.

Indeed, and I do like the markers idea a lot. But the reason the nvram variable check was introduces is specifically to support the mount --bind. A browser would not know what to do with those markers on the page and extra code is processed, often braking the page visualisation.

The only alternative I can think about really is to introduce a JS function to map HW models to chipset and do an interpretation of the compiling markers, but somehow this is more difficult and complex of the current solution...






> What I meant, now we have mix of both - you remove some of `EXT-SW-...` but not all of them. So let's use them as before.

Let me cross-check the markers/nvram logic and see if I can come up with something that has lower impact

pedro0311:
I have corrected this PR (available here: https://freshtomato.org/downloads/freshtomato-arm/2025/2025.5/TESTS/ethernet.patch), but honestly, I don't like it: ethernet.svg takes a very long time to respond to the options passed there, and sometimes the options are changed incorrectly (e.g. the LEDs don't blink, or blink in the wrong color), and sometimes an ethernet image can disappear and not be displayed anymore.

wellloaded:
> I have corrected this PR (available here: https://freshtomato.org/downloads/freshtomato-arm/2025/2025.5/TESTS/ethernet.patch), but honestly, I don't like it: ethernet.svg takes a very long time to respond to the options passed there, and sometimes the options are changed incorrectly (e.g. the LEDs don't blink, or blink in the wrong color), and sometimes an ethernet image can disappear and not be displayed anymore.

Hum... I didn't experience this on my side, otherwise I would not have created the PR. What browser are you using?
I'm wondering if there's a quick fix for this issue. The Ethernet icons are passed as objects so that caching can take effect, so I would expect a parameter passed to it to be fed to the cache and be quick. Either there's an issue with the way the parameters are passed, or perhaps some minor issue with the logic of the SVG perhaps? I guess adding some logs for troubleshooting might help.

pedro0311:
I checked: Pale Moon, Chrome, Opera - OK
FF - problems I described earlier.

wellloaded:
> What I meant, now we have mix of both - you remove some of `EXT-SW-...` but not all of them. So let's use them as before.

I went through the diff and I can't see any ETX-SW removed. Perhaps the content was replaced that yes, but I don't see any removed. Admittedly, there are lots of changes in this page, hopefully all for the best.

The target idea about nvram checks and compiling markers here is:

A) When the page is used as "mount --bind", the nvram code takes precedence and allows the correct visualisation of the page

B) When compiling though, those nvram checks should be excluded from the firmware so, compiling markers take precedence.

So A) is completed where B) still requires the REMOVE-BEGIN and REMOVE-END markers to complete the job. Check this version in attachment, it takes the idea of the embedded SVG but to have a centralised icon this is placed in the tomato.js now (could be a different mapped .js if needed).

[vlan_JSicon.zip](https://github.com/user-attachments/files/25186367/vlan_JSicon.zip)

wellloaded:
> I checked: Pale Moon, Chrome, Opera - OK FF - problems I described earlier.

For me to better understand, is the FF issue experienced on the Overview or on the VLAN page? Or both?

pedro0311:
You can't create an ASP page that's correct in both cases (bind vs. compilation) – you'll always end up with redundant code that will obfuscate the code and be completely unnecessary during proper compilation. That's why I'm against such solutions. For testing, simply comment out the code between `EXT-SW-BEGIN -END` because I don't think anyone is using an Asus RT-AC88U for testing.

Also use the complete code for testing - do not upload the SVG icon with support to tomato.js, because later - when you move this code to the correct .js file - it may turn out that not everything works correctly/identically.

Also: have you checked (e.g., with AI assistance) which browsers support such interactive SVG icons? It may turn out that half of users will be left with broken ethernet icons...

> For me to better understand, is the FF issue experienced on the Overview or on the VLAN page? Or both?

Both. The problem occurs on FF 115.32.0esr (64) but on the latest version everything seems to work more or less.

wellloaded:
> You can't create an ASP page that's correct in both cases (bind vs. compilation) – you'll always end up with redundant code that will obfuscate the code and be completely unnecessary during proper compilation. That's why I'm against such solutions. For testing, simply comment out the code between `EXT-SW-BEGIN -END` because I don't think anyone is using an Asus RT-AC88U for testing.

The approach I took, and it seems to be working (so far at least), is to relay on nvram to direct the correct display of code based on the nvram model when this is **mounted --bind**. Those extra nvram checks are within REMOVE tags, so they will not end up in the compiled version. The compiler marks we already know the will already include/exclude what's needed based on the model. As a logic, it works for me, the issue @HommeOursPorc reported is on the overview that's where the double nvram+compiling conditions have not been added, but for VLAN and Virtual Wireless we are able to mount the page and use the controls like if they were compiled in. But perhaps you have a better view on any issue I'm not considering?

e.g.:
```
/* EXTSW-BEGIN */
/* REMOVE-BEGIN */
		if (hasExtSw) {
/* REMOVE-END */
			cols.push({ type: 'select', options: portOptions, prefix: '<div class="centered">', suffix: '<\/div>' });
/* REMOVE-BEGIN */
		}
/* REMOVE-END */
/* EXTSW-END */
```


> 
> Also use the complete code for testing - do not upload the SVG icon with support to tomato.js, because later - when you move this code to the correct .js file - it may turn out that not everything works correctly/identically.

For mounted files (without going compiling the test I mean) we must relay on an existing .js e.g. placing it into /var/wwwext/ and sourcing it as ext/ethernet-icon.js will not work (JS limitation). It's not my intention to leave it in the tomato.js it will be a new external file as mentioned in the forum as well. I will of course perform a compilation test before any PR.


> 
> Also: have you checked (e.g., with AI assistance) which browsers support such interactive SVG icons? It may turn out that half of users will be left with broken ethernet icons...

One issue I have found and I'm working hard on this one, is that once in a blue moon one or more icon go from e.g. "Unplugged" to "UNSUPPORTED". I'm really having hard time understanding why as this appens but everythign usggests there's an issue somewhere with the backend passing the correct parameters. AI suggested to cache the last well know result in a variable and skip any temporary UNSUPPORTED state. It seems to me overall a reasonable suggestion. I'm testing this little caching function as I speak.



> > For me to better understand, is the FF issue experienced on the Overview or on the VLAN page? Or both?
> 
> Both. The problem occurs on FF 115.32.0esr (64) but on the latest version everything seems to work more or less.

Hopefully the move from <Object> to <embedded> and the little caching code I'm referring to above are a fix for this visualisation glitch. Fingers crossed. :-p

Thanks

HommeOursPorc:
> One issue I have found and I'm working hard on this one, is that once in a blue moon one or more icon go from e.g. "Unplugged" to "UNSUPPORTED". I'm really having hard time understanding why as this appens but everythign usggests there's an issue somewhere with the backend passing the correct parameters. AI suggested to cache the last well know result in a variable and skip any temporary UNSUPPORTED state.

I've seen a somewhat similar issue in Tomato64, but the port were tagged as UNKNOWN, not UNSUPPORTED. After a bit of digging, I found the thread I made about this:
https://www.linksysinfo.org/index.php?threads/tmp-ethernet-state-unknown.78745/

I didn't check further, but if it is a similar issue, the port state might come from: 
https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/others/ethstate

So it use `robocfg showports` to get the port info. Tomato64 doesn't have robocfg, but the command that T64 use is the one reporting the UNKNOWN. So maybe the message comes from robocfg???

No idea and I didn't investigate. It just made me think of this.

pedro0311:
https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/misc.c#L533
-->
https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/others/ethstate

wellloaded:
> https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/misc.c#L533 --> https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/others/ethstate

If I had to guess may be the page reloads while the file is being re-created?

pedro0311:
What file?
The asp page calls (using ajax) the C function asp_etherstates(), which - using /usr/sbin/ethstate -> robocfg - prints the port state on this page.

I wonder if changing from system() to eval() will help here.
https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/misc.c#L543

wellloaded:
> What file? 

The /tmp/ethernet.state 

Although, reading carefully @HommeOursPorc post here:
https://www.linksysinfo.org/index.php?threads/tmp-ethernet-state-unknown.78745/

It seems like the file itself is populated with unwanted content at times, so there must be something else under the bonnet instead.

pedro0311:
@wellloaded : I have a fix for VLANS: move `show()` from `earlyInit()` to `init()` .
Regarding Overview : still thinking...

wellloaded:
> @wellloaded : I have a fix for VLANS: move `show()` from `earlyInit()` to `init()` . Regarding Overview : still thinking...

I'll try. I suppose that applied to any page with icons (so the overview as well) right? Why would overview be different?

pedro0311:
This fix (for FF 115.32.0esr (64)) is only for VLANs.
The Overview page is built differently.

// EDIT: ...and moving `show()` to `init()` there didn't help.

wellloaded:
> This fix (for FF 115.32.0esr (64)) is only for VLANs. The Overview page is built differently.
> 
> // EDIT: ...and moving `show()` to `init()` there didn't help.

Hum... I did that, but I still see the issue on the VLAN page.

pedro0311:
> Hum... I did that, but I still see the issue on the VLAN page.

What issue is still relevant?

BTW, you may move `show()` to `init()` also in Overview - this won't solve the issue with "empty" requests and responses when first loading a page in this version of FF (like it helped with the VLANs page), but it can be done for consistency.

wellloaded:
> What issue is still relevant?

The connected icons on the vlan page disappear now and again. This regardless of the show() being under earlyInit() or init(). e.g. I don't see any difference in behaviour moving it. FF 147.0.3 (64-bit)

 
> BTW, you may move `show()` to `init()` also in Overview - this won't solve the issue with "empty" requests and responses when first loading a page in this version of FF (like it helped with the VLANs page), but it can be done for consistency.

Ok.

pedro0311:
Try this: https://github.com/FreshTomato-Project/freshtomato-arm/commit/0d57069bfa2efe3cbec47df3d938f0259ec05a3c

pedro0311:
> this won't solve the issue with "empty" requests and responses when first loading a page in this version of FF

I mean this:

Clipboard01

PR #191, opened 2026-01-04
GUI: tools-survey.asp - upgrade to v2.01

wellloaded:
- Changed charts from canvas to SVG
- Sped up the initial display of data after initial page load, this might display 2.4Ghz first and 5Ghz after 1 or 2 seconds, but it's faster than waiting the refresh timer
- Fixed some odd loading behaviour, especially when the page first loads
- "error: unable to fetch scan results" is not displayed any more, if no data is available at refresh it will just keep trying until a first result are obtained
- Added Loading... + spinner to the charts at page load to suggest data is coming
- “Display” filter is hidden if only one frequency is enabled under basic/network
- Fixed first page visit ever when a refresh is not yet set (now defaults to 15s)
- Fixed bug when SSIDs contains the character: "
- Overall better look and feel using the page

### DIFF
--- a/release/src-rt-6.x.4708/router/www/tools-survey.asp
+++ b/release/src-rt-6.x.4708/router/www/tools-survey.asp
@@ -37,8 +37,7 @@ function internalWiFI(interface) {
 	return new Promise(function(resolve, reject) {
 		var cmd = new XmlHttp();
 		cmd.onCompleted = function(text, xml) {
-			var result = text.replace(/\x0a/g, ' ');
-			resolve(result);
+			resolve(text);
 		};
 		cmd.onError = function(x) {
 			var error = 'ERROR: '+x.replace(/\x0a/g, ' ');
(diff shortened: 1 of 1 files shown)


HommeOursPorc:
Is it only me or copilot is annoying!? Is that a new option or it is forced?

wellloaded:
I think it's something on my side to be fair. I need to tweak it to give local feedback only and not after a PR is created.

scoobasteve83:
> Is it only me or copilot is annoying!? Is that a new option or it is forced?

:) now now you are not treating it strictly equal.  

Ehhh it has its purpose to be honest likely brought up a good point or two. I'm all for making code more easily readable through comments and better names especially in a codebase this large no?

pedro0311:
@wellloaded : is it ready for a review?

wellloaded:
> @wellloaded : is it ready for a review?

Yes

pedro0311:
OK

pedro0311:
Have you checked how the graphs behave when we disable one 5GHz WL in routers with two 5GHz WLs? 

I've had reports that in your previous version, the 5GHz graph wouldn't display at all in that case (and I can't check it right now...)

wellloaded:
> Have you checked how the graphs behave when we disable one 5GHz WL in routers with two 5GHz WLs?
> 
> I've had reports that in your previous version, the 5GHz graph wouldn't display at all in that case (and I can't check it right now...)

So the PR is about 4 months old now, I honestly don't remember the little details any longer. I do remember having tested this on my RT-AC3200 with 1x 2.4GHz and 2x5GHz so I assume the answer is 'yes', it will work.

What I can tell you is that a spotted a number of minor issues throughout time while using the old version, so those cases were identified and corrected. There *might* be something unfixed based on particular user scenario (which we can fix when reported) but there are so many wireless modes I can't possibly test all the scenarios. So take it as: I'm sure this version works better than the prev version, it's a fact, but if anything is reported back by the users we can address one by one.
Let's remember this is a non-critical page.

pedro0311:
OK< thanks!

pedro0311:
https://github.com/FreshTomato-Project/freshtomato-arm/commit/8106f1dfbdfe077bd6ea13d907073cbce85c4841

PR #193, opened 2026-01-05
GUI: advanced-misc.asp - Ethernet Port Health

wellloaded:
Adds a new function to monitor errors on the Ethernet ports. It prevents the router from crashing when having to deal with numerous bad frames, whether related to port speed mismatch or cabling issues. It uses a new, single, porthealth_cfg nvram variable,

image

It can monitor (only), try to recover or disable when threshold is breached.
When enabled it runs every minute, however its execution time is around 0.02s-0.03s.
The output is not human friendly but can be verified under /tmp/porthealth/

### DIFF
--- a/release/src-rt-6.x.4708/router/others/Makefile
+++ b/release/src-rt-6.x.4708/router/others/Makefile
@@ -8,6 +8,9 @@ install:
 	install -d $(INSTALLDIR)/usr/bin
 	install -d $(INSTALLDIR)/lib
 
+# Port health monitor/recover script
+	install -m 0700 porthealth.sh $(INSTALLDIR)/usr/sbin/porthealth.sh
+
 ifeq ($(TCONFIG_USB),y)
 # Optware perl symlink
 	install -d $(INSTALLDIR)/usr/bin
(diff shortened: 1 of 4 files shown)


scoobasteve83:
is there a sick face that shows up in place of the port or red exclamation point? Or what happens exactly to indicate the results other than say maybe being a lower speed connection potentially. Cool idea. You've had some bangers lately. Oh I've looked it up it don't appear that it does anything to show port issues detected there are 4pages where code is change and overview is not one of them so I can't see how it would but you should still be able to see degraded speed.

pedro0311:
@wellloaded : should we also add it to MIPS with 4MB flash? What do you think?

wellloaded:
> @wellloaded : should we also add it to MIPS with 4MB flash? What do you think?

I believe it's better to keep this ARM only, that's where it was developed. We go fetching parameters from the /proc, and I fear MIPS (even R2) will have different considerations to be taken into account, not to mention the execution time (considering it's every minute) it's going to raise significantly from the current 0.02-0.03s.

pedro0311:
OK, thanks

pedro0311:
Added as a separate commit, because of some cosmetic: https://github.com/FreshTomato-Project/freshtomato-arm/commit/e90e7553ab321bbdf633972f98e329743da9d387

PR #214, opened 2026-01-27
Update defaults.c disable telnet enable at startup

scoobasteve83:
[ $1 -ge 20 ] && telnetd -p (yourporthere) -l /bin/sh

already exists to bypass password...

Don't this open up telnet on port (yourporthere) and drop you into a root shell where you can already get the username and password if you are locked out using nvram get and the variables?

### DIFF
--- a/release/src-rt-6.x.4708/router/shared/defaults.c
+++ b/release/src-rt-6.x.4708/router/shared/defaults.c
@@ -1127,7 +1127,7 @@ struct nvram_tuple router_defaults[] = {
 	{ "ttb_url",			"http://ttb.mooo.com http://ttb.ath.cx http://ttb.ddnsfree.com", 0 },	/* Tomato Themes Base - default URL */
 #endif
 	{ "web_svg",			"1"				, 0 },
-	{ "telnetd_eas",		"1"				, 0 },
+	{ "telnetd_eas",		"0"				, 0 },
 	{ "telnetd_port",		"23"				, 0 },
 	{ "sshd_eas",			"1"				, 0 },	/* enable sshd by default */
 	{ "sshd_pass",			"1"				, 0 },


HommeOursPorc:
You need to press the button for this, so you need physical access to the router in the first place.

Screenshot_20260126_192549_Firefox

scoobasteve83:
precisely... what is the point physical access to start telnet yes you can always start the server as well with the GUI... That is kind of how you should have things by default don't you think?

HommeOursPorc:
It is meant to be a emergency access

scoobasteve83:
Of course emergency access is there with a button press for 20 seconds. Yes. Or you mean having it always on by default?

scoobasteve83:
I left the emergency button press alone :) only disabled enable at startup changed a 1 to a 0 don't worry I think the button press is useful yes....

Or did I delete it by mistake errr I better look now seeing as how you seem worried I did highlight it when I was looking at the references to telnet in code when looking for telnet references I suppose its possible ehhhh :/

EDIT: Nope I double checked I left the button press alone.

scoobasteve83:
hrmmm alternatively add box if you want it to run 24/7 uncheck the box real simple like the attack surface would be greatly reduced I think Fresh Tomato would be much safer with this at a minimum.

admin-access.asp

```
createFieldTable('', [
    { title: 'Auto-disable after 10m', name: 'f_auto_disable_telnet', type: 'checkbox', value: nvram.auto_disable_mgmt == '1' }
]);
```
rc/services.c
add the following
```
if (nvram_get_int("auto_disable_telnet")) {
    // Run a background shell command to wait 10 minutes then stop services
    system("(sleep 600; service telnetd stop; nvram set telnet_eas=0; nvram commit) &");
}
```

defaults.c

`{ "auto_disable_telnet",		"1"				, 0 }, /* enable telnet on 10min timer by default */`

If we do it this way we should add a box for SSH as well and have that disabled after 10mins by default.... But for sure telnet don't you think IDK.... 

I think it's a good precaution if you think we should leave it enabled by default.

pedro0311:
Yes, it should be done long ago. Thanks.

Issue #155, opened 2025-10-23
Dynamic DNS stopped updating (continuation of #80)


quickishfm:
Hi,
#80 has the main issue but i am on 2025.2 and still have this issue where the DDNS has stopped updating after 105 days uptime (currently at 131 days). To include information from my latest comment on the closed #80 issue;

I opened an SSH connection and noticed that the lockfile has been open since 10 minutes since the last update, which co-incides with the WAN IP checker interval.

root@unknown:/# find -iname "*ddns*"
./sbin/ddns-update
./tmp/cru-ddns
./tmp/var/lock/ddns0.lock
./tmp/var/lock/ddns3.lock
./tmp/var/lock/ddns2.lock
./tmp/var/lock/ddns1.lock
./tmp/var/lib/mdu/ddnsx3.msg
./tmp/var/lib/mdu/ddnsx2.msg
./tmp/var/lib/mdu/ddnsx1.msg
./tmp/var/lib/mdu/ddnsx0.extip
./tmp/var/lib/mdu/ddnsx0.conf
./tmp/var/lib/mdu/ddnsx0.msg
./tmp/var/lib/mdu/ddnsx0.cache
./www/basic-ddns.asp
root@unknown:/# ls -al /tmp/var/lock
drwxrwxrwx    2 root     root           260 Oct 13 08:45 .
drwxr-xr-x   11 root     root           220 Jan  1  1970 ..
-rw-rw-rw-    1 root     root             4 Jan  1  1970 action
-rw-------    1 root     root             0 Sep 28 23:59 cron.lock
-rw-------    1 root     root             0 Sep 28 23:59 ddns0.lock
-rw-------    1 root     root             0 Jun 14 15:39 ddns1.lock
-rw-------    1 root     root             0 Jun 14 15:39 ddns2.lock
-rw-------    1 root     root             0 Jun 14 15:39 ddns3.lock
-rw-------    1 root     root             0 Jan  1  1970 firewall.lock
-rw-r--r--    1 root     root             0 Jun 14 15:39 httpd.lock
-rw-------    1 root     root             0 Oct 13 08:45 restrictions.lock
-rw-r--r--    1 root     root             0 Jan  1  1970 usb.lock
-rw-------    1 root     root             0 Jan  1  1970 xtables.lock
root@unknown:/# 

Is there any more debugging I can do to find why this keeps happening? I have not rebooted the router yet so I can still live debug.

quickishfm:
[...]

PR #216, opened 2026-01-29
Arm master freshtomato updates

MarsTomato:
updates, checked both - working correct

### DIFF
--- a/release/src-rt-6.x.4708/Makefile
+++ b/release/src-rt-6.x.4708/Makefile
@@ -150,6 +150,48 @@ else
 	current_V2 = ""
 endif
 
+# Macro to inject firmware name and package the firmware (Default: Asus Header)
+define CREATE_INJECT_MODEL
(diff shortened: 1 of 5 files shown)


pedro0311:
Would you be so kind to add to 1st commit name (ntpd) at the end: " (close #155)" ? Thanks!

I mean something like:
"ntpd: increase limits (Max Memory & Max Processes) (close #155)"

MarsTomato:
> Would you be so kind to add to 1st commit name (ntpd) at the end: " (close #155)" ? Thanks!
> 
> I mean something like: "ntpd: increase limits (Max Memory & Max Processes) (close #155)"

i hesitated to do that, because maybe there is still a problem for this user in very rare conditions? no further feedback so far.
thats why i did write "fix/improvement"
But i can close it

pedro0311:
No, no I 'll merge it.

PR #217, opened 2026-01-30
build: move firmware name injection to /rom/filename/

MarsTomato:
Update/improve commit https://github.com/FreshTomato-Project/freshtomato-arm/commit/02fcac338d79003649ac2afac6ba377afedf04cc

### DIFF
--- a/release/src-rt-6.x.4708/Makefile
+++ b/release/src-rt-6.x.4708/Makefile
@@ -153,15 +153,15 @@ endif
 # Macro to inject firmware name and package the firmware (Default: Asus Header)
 define CREATE_INJECT_MODEL
 	# 1. Prepare environment: Create target directory and remove old artifacts
-	mkdir -p router/arm-uclibc/target
-	rm -f router/arm-uclibc/target/*.trx router/arm-uclibc/target/*.chk
+	mkdir -p router/arm-uclibc/target/rom/filename
+	rm -rf router/arm-uclibc/target/rom/filename/*
 	rm -f image/linux-lzma.trx
 
 	# 2. Sanitize the model name: Remove quotes and store in a variable
 	$(eval SAN_IMAGE := $(shell echo $(1) | tr -d '"' | tr -d "'"))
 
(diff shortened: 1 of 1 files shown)


wellloaded:
@MarsTomato if this works /rom i indeed a better location. This said can you please make 100% certain this doesn't brake the trx_asus.c logic? As mentioned elsewhere Asus (only) performs a second CRC calculation. So worth testing "a" Asus build (any) with this new /rom location. If it doesn't work while upgraded manually you should be getting a CRC error, after a short countdown, on the manual upgrade page after you clock "Upgrade"

MarsTomato:
> @MarsTomato if this works /rom i indeed a better location. This said can you please make 100% certain this doesn't brake the trx_asus.c logic? As mentioned elsewhere Asus (only) performs a second CRC calculation. So worth testing "a" Asus build (any) with this new /rom location. If it doesn't work while upgraded manually you should be getting a CRC error, after a short countdown, on the manual upgrade page after you clock "Upgrade"

It is safe because the checksums are calculated after the changes. We do the following:

1) add the folder / file

2) mksquashfs / pack

3) trx and trx_asus calculate the final container checksums

"sealing" happens last, the final image remains perfectly consistent

To double check i tested it with my Asus RT-N18U - NO issuses / CRC OK :
```
root@RT-N18U-Mars:/# ls
bin       cifs2     home      mmc       opt       root      tftpboot  var
bkp       dev       jffs      mnt       proc      sbin      tmp       www
cifs1     etc       lib       nas       rom       sys       usr
root@RT-N18U-Mars:/# cd rom
root@RT-N18U-Mars:/rom# ls -1
dot-servers.dat
etc
filename
root@RT-N18U-Mars:/rom# cd filename/
root@RT-N18U-Mars:/rom/filename# ls -1
freshtomato-RT-N18U-K26ARM-2025.5-07012026v003-VPN-64K-NOSMP.trx
root@RT-N18U-Mars:/rom/filename#
```

pedro0311:
Thanks!

PR #220, opened 2026-02-07
rc: wireguard.c: fix concurrency issues

lancethepants:
- Replace exit(0) with _exit(0) in forked child.
- Run all dnsmasq/firewall restarts through the init process to prevent race conditions.
- Guard all firewall operations with locks.


lancethepants:
Rebooting many times on occasion either wireguard or dnsmasq would simply not be running. Forcing start-ups/restarts to go through pid 1 should eliminate race conditions.

pedro0311:
Thank you.
I assume, exit() should be changed to _exit() also in mysql.c, nginx.c and transmission.c

PR #221, opened 2026-02-07
tomato.js - allows for placeholder to work on password fields

wellloaded:
Preserve placeholder attribute when swapping input types.


pedro0311:
Thanks!

PR #226, opened 2026-02-17
rc: add NVRAM variable migration for renamed OpenVPN settings on firmware upgrade

lancethepants:


lancethepants:
Tested on R7000, preserved OpenVPN config when upgraded from 2025.4 release. Similar to RMerlin AsusWrt method.

HommeOursPorc:
Are the old variable currently flushed when booting 2026.1 for the first time? I have run @pedro0311 script, then I worked on my ftup script and didn't upgrade right away. I don't remember if I rebooted, I think I did reboot somewhere between executing the script and finally upgrade the router... So I thought the old values would have been brought back by defaults.c , but when looking into the nvram dump of 2026.1 after the upgrade, the old values where gone. 

So either they got flushed at boot time... Or I did not reboot between the script execution and the 2026.1 upgrade.... I just can't remember, but if they are flushed at boot, isn't it kind of too late? Doesn't this would have been needed before releasing 2026.1?

Just some thoughts... I can't remember what I did exactly, so....

lancethepants:
With the script the old nvram variables are unset (removed), and defaults.c was also changed in the same commit so it would not restore old values.

HommeOursPorc:
Ok I just tested it. My assumptions were correct, not upgrading right away and rebooting in between running the script and upgrading result in the old nvram variables being reinstated by defaults.c. The old variable are not removed when booting 2026.1.

Finally it look like my memory failed me and that I did not reboot in between the script execution and the upgrade. I just tested it by downgrading and re-upgrading and now I have all the old and new variables in my dump 😆

```bash
root@AC68U-Relais:/tmp/home/root# nvram show 2>/dev/null | grep -c ^vpns
84
root@AC68U-Relais:/tmp/home/root# nvram show 2>/dev/null | grep -c ^vpn_server
84
```

Edit: I mean I
Downgraded to 2025.5.
Run Pedro's script.
Reboot.
After reboot the old variable are restored by the firmware. 
Upgraded back to 2026.1.
Now I have both the old variable and new ones.

pedro0311:
@lancethepants : I'm just curious: what's wrong with this PR?

lancethepants:
ah, I thought I double submitted the other PR, this one is good.

pedro0311:
Thanks!

FYI: with this commit, the MIPS MiniVPN image has 3862580 bytes, so we are already close to the 3866624 bytes limit (that's why I solved the problem with an sh file).

PR #227, opened 2026-02-21
GUI: fix id mismatch in generation of vpn client config

suciua:
Fix for the generation of VPN client config failing because of mismatching IDs since renaming the nvram variables. The HTML contains a field called `vpns1_gen_progress_div` but the function `downloadClientConfig` tries to find the selector `server1_gen_progress_div` instead. The element cannot be found which is causing a JavaScript exception.

### DIFF
--- a/release/src-rt-6.x.4708/router/www/vpn-server.asp
+++ b/release/src-rt-6.x.4708/router/www/vpn-server.asp
@@ -535,10 +535,10 @@ function downloadClientConfig(num) {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


suciua:
I didn't know how to test it other than renaming the field ID back to `server1_gen_progress_div` in the browser and verifying the the generation works.

pedro0311:
Full fix: https://github.com/FreshTomato-Project/freshtomato-arm/pull/228

PR #231, opened 2026-02-26
rc: Add trx, chk and bin firmware images integrity validation

HommeOursPorc:
Hi Pedro! 

This is AI generated and just before making this PR I noticed the AI messed things up a little bit. I will post this anyway since it is late and worked too much on this.

This validate image files before flashing them.

What work:
  - It seem to correctly check images files in trx, chk and bin format.
  - It can output a simple 1 line valid output 
  - Or with the -v option it can give details. 
  - It can take many files as input and check then all at once.
  - option -h to get help. 
  - option -v must come before -i

What doesn't work: 
  - If the image is invalid there's no output and no error code. echo $? Return 0 which is NOT ok.
  - option --help have probably been removed by the Artificial Dumb Dumb.

I can try to fix this tomorrow or maybe it is an easy fix??? Or maybe this is a waste of time? 

```bash
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -h
TRX/CHK/BIN Firmware Checker
Usage: trx_check [-v] -i <file1> [file2...]

# broken image make no output and return a 0 error code. NOT good!
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -i test.trx
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# echo $?
0
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -v -i test.trx

root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -v -i vpn.trx

Analyzing File: vpn.trx
Size: 15577088 bytes
Type: TRX (Broadcom)
  Magic: 0x30524448 (HDR0)
  Header Length: 15577088
  Expected CRC:   0x423fd288
  Calculated CRC: 0x423fd288
Final Result: ✓ VALID
------------------------------

root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -i vpn.trx
vpn.trx: VALID

root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -i vpn.trx aio.trx
vpn.trx: VALID
aio.trx: VALID

root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -v -i vpn.trx aio.trx

Analyzing File: vpn.trx
Size: 15577088 bytes
Type: TRX (Broadcom)
  Magic: 0x30524448 (HDR0)
  Header Length: 15577088
  Expected CRC:   0x423fd288
  Calculated CRC: 0x423fd288
Final Result: ✓ VALID
------------------------------

Analyzing File: aio.trx
Size: 31719424 bytes
Type: TRX (Broadcom)
  Magic: 0x30524448 (HDR0)
  Header Length: 31719424
  Expected CRC:   0x1f37b957
  Calculated CRC: 0x1f37b957
Final Result: ✓ VALID
------------------------------
```

What do you guys think? I don't know if you all saw my other iteration, but the first one by Copilot using Claude was about 400 lines. This one by Gemini is ~150 lines.

IMO this is worth it. You guys tell me.

On this, good night for me and probably good morning to you on the other side of the Atlantic 🙂

### DIFF
--- a/release/src-rt-6.x.4708/router/rc/Makefile
+++ b/release/src-rt-6.x.4708/router/rc/Makefile
@@ -28,7 +28,7 @@ OBJS = rc.o init.o interface.o network.o wan.o services.o dhcp.o \
        firewall.o ppp.o telssh.o wnas.o listen.o redial.o \
        led.o qos.o forward.o misc.o mtd.o buttons.o restrict.o \
        gpio.o sched.o bwlimit.o arpbind.o tomatoanon.o mwan.o pbr.o \
-       dnsmasq.o format.o
+       dnsmasq.o format.o trx_check.o
 
 ifeq ($(or $(TCONFIG_BCMARM),$(TCONFIG_BLINK)),y)
  OBJS += blink.o blink_br.o
@@ -177,6 +177,7 @@ endif
 	@cd $(INSTALLDIR)/sbin && ln -sf rc ppp_event
 
 	@cd $(INSTALLDIR)/sbin && ln -sf rc ntpd_synced
+	@cd $(INSTALLDIR)/sbin && ln -sf rc trx_check
 
 ifeq ($(TCONFIG_BCMARM),y)
 	@cd $(INSTALLDIR)/sbin && ln -sf rc mtd-write2
--- a/release/src-rt-6.x.4708/router/rc/rc.c
+++ b/release/src-rt-6.x.4708/router/rc/rc.c
@@ -18,6 +18,8 @@
  #include <setjmp.h>
 #endif
 
+extern int trx_check_main(int argc, char **argv);
+
 /* needed by logmsg() */
 #define LOGMSG_DISABLE	DISABLE_SYSLOG_OSM
 #define LOGMSG_NVDEBUG	"rc_debug"
@@ -247,7 +249,7 @@ void run_del_firewall_script(const char *infile, char *outfile)
  * @param	out	pointer to a buffer where the function writes the normalized result
  * @param	outlen	size of the out buffer
  * @return	1 if str is a valid IPv4 or IPv4/mask; out receives a normalized string in the form A.B.C.D/n with n coerced into semantics, defaulting to /32 when absent or invalid
- *		2 if str is a valid IPv4 range “A.B.C.D-E.F.G.H” with no internal whitespace and both addresses in the same /24; out receives the unchanged “A.B.C.D-E.F.G.H”
+ *		2 if str is a valid IPv4 range ï¿½A.B.C.D-E.F.G.Hï¿½ with no internal whitespace and both addresses in the same /24; out receives the unchanged ï¿½A.B.C.D-E.F.G.Hï¿½
  * 		0 if str is something else (FQDN?)
  */
 static int check_string(const char *str, char *out, size_t outlen)
@@ -894,6 +896,7 @@ static const applets_t applets[] = {
 #ifdef TCONFIG_ROAM
 	{ "roamast",			roam_assistant_main		},
 #endif
+    { "trx_check", trx_check_main },
 	{NULL, NULL}
 };
 
--- a/release/src-rt-6.x.4708/router/rc/trx_check.c
+++ b/release/src-rt-6.x.4708/router/rc/trx_check.c
@@ -1,0 +1,147 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <stdint.h>
+#include <sys/stat.h>
+#include <sys/mman.h>
+#include <fcntl.h>
+#include <unistd.h>
+#include <stddef.h>
+
+#define TRX_MAGIC           0x30524448  /* "HDR0" */
+#define NETGEAR_MAGIC       0x5E24232A  /* Netgear CHK */
+#define TRX_MAX_LEN         0x4000000   /* 64MB */
+
+static uint32_t crc32_tab[256];
+static int verbose = 0;
+
+static void init_crc32(void) {
+    uint32_t i, j, c;
+    for (i = 0; i < 256; i++) {
+        for (c = i, j = 8; j > 0; j--)
+            c = (c & 1) ? (0xEDB88320L ^ (c >> 1)) : (c >> 1);
+        crc32_tab[i] = c;
+    }
+}
+
+static uint32_t crc32_calc(uint32_t crc, const uint8_t *buf, size_t len) {
+    while (len--) crc = crc32_tab[(crc ^ *buf++) & 0xFF] ^ (crc >> 8);
+    return crc;
+}
+
+struct trx_header {
+    uint32_t magic, len, crc32, flag_version, offsets[3];
+};
+
+static int perform_check(const char *fn) {
+    int fd, res = 1;
+    struct stat st;
+    uint8_t *buf;
+
+    if (!fn) return 1;
+    fd = open(fn, O_RDONLY);
+    if (fd < 0) {
+        fprintf(stderr, "ERROR: Cannot open %s\n", fn);
+        return 4;
+    }
+    
+    if (fstat(fd, &st) < 0 || st.st_size < 32) {
+        close(fd);
+        return 3;
+    }
+
+    buf = mmap(NULL, st.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
+    if (buf == MAP_FAILED) {
+        close(fd);
+        return 4;
+    }
+
+    if (verbose) {
+        printf("\nAnalyzing File: %s\n", fn);
+        printf("Size: %zu bytes\n", (size_t)st.st_size);
+    }
+
+    uint32_t magic = *(uint32_t *)buf;
+
+    /* 1. TRX Check */
+    if (magic == TRX_MAGIC) {
+        struct trx_header *h = (struct trx_header *)buf;
+        if (verbose) {
+            printf("Type: TRX (Broadcom)\n");
+            printf("  Magic: 0x%08x (HDR0)\n", h->magic);
+            printf("  Header Length: %u\n", h->len);
+        }
+
+        if (h->len > st.st_size || h->len > TRX_MAX_LEN) {
+            if (verbose) printf("  ERROR: Length mismatch or exceeds 64MB limit\n");
+            res = 3;
+        } else {
+            init_crc32();
+            uint32_t calc = crc32_calc(0xFFFFFFFF, (uint8_t *)&h->flag_version, h->len - 12);
+            if (verbose) {
+                printf("  Expected CRC:   0x%08x\n", h->crc32);
+                printf("  Calculated CRC: 0x%08x\n", calc);
+            }
+            res = (calc == h->crc32) ? 0 : 2;
+        }
+    } 
+    /* 2. CHK Check */
+    else if (magic == NETGEAR_MAGIC) {
+        if (verbose) printf("Type: Netgear CHK (Verified by Magic)\n");
+        res = 0; 
+    } 
+    /* 3. BIN/Generic Check */
+    else {
+        int found = 0;
+        for (int i = 0; i < 512 && i < st.st_size - 32; i++) {
+            if (memcmp(buf + i, "U2ND", 4) == 0 || memcmp(buf + i, "HDR0", 4) == 0) {
+                if (verbose) printf("Type: BIN (Linksys/ASUS) - Identifier found at offset %d\n", i);
+                found = 1;
+                res = 0;
+                break;
+            }
+        }
+        if (!found && verbose) printf("Type: Unknown (Magic: 0x%08x)\n", magic);
+    }
+
+    /* Final Output */
+    if (!verbose) {
+        printf("%s: %s\n", fn, (res == 0) ? "VALID" : "INVALID");
+    } else {
+        printf("Final Result: %s\n------------------------------\n", (res == 0) ? "✓ VALID" : "✗ INVALID");
+    }
+
+    munmap(buf, st.st_size);
+    close(fd);
+    return res;
+}
+
+int trx_check_main(int argc, char *argv[]) {
+    int ch, i;
+    int first_file_idx = -1;
+    verbose = 0;
+    optind = 0;
+
+    while ((ch = getopt(argc, argv, "vhi:")) != -1) {
+        switch (ch) {
+            case 'v': verbose = 1; break;
+            case 'i': first_file_idx = optind - 1; break;
+            case 'h': 
+            default:
+                printf("TRX/CHK/BIN Firmware Checker\n");
+                printf("Usage: trx_check [-v] -i <file1> [file2...]\n");
+                return 0;
+        }
+    }
+
+    if (first_file_idx != -1) {
+        for (i = first_file_idx; i < argc; i++) {
+            if (strcmp(argv[i], "-i") == 0) continue;
+            perform_check(argv[i]);
+        }
+    } else {
+        printf("Error: Use -i to specify firmware files.\n");
+    }
+
+    return 0;
+}


HommeOursPorc:
Forgot to put CHK and BIN output: 
```
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -v -i ft.bin ft.chk

Analyzing File: ft.bin
Size: 8070144 bytes
Type: BIN (Linksys/ASUS) - Identifier found at offset 14
Final Result: ✓ VALID
------------------------------

Analyzing File: ft.chk
Size: 10039354 bytes
Type: Netgear CHK (Verified by Magic)
Final Result: ✓ VALID
------------------------------
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# trx_check -i ft.bin ft.chk
ft.bin: VALID
ft.chk: VALID
```

HommeOursPorc:
Hmmm, I think it doesn't do any CRC check on chk and bin ... ?? But I think there's one?? Ahhhhhh. I think we can close this. Back to the drawing board.

HommeOursPorc:
BTW, is there a faster way than to compile the whole firmware and flash the router each time I change trx_check.c 

It is past 3am here. Time for bed! 

But judging by Gemini, this is the greatest trx_check of the history of trx_check. It is the best and it is all because of Gemini, otherwise it wouldn't have happened. Gemini made trx_check great again.

```c
#include <stdio.h>
#include <stdlib.h>
#include <string.h>
#include <stdint.h>
#include <sys/stat.h>
#include <sys/mman.h>
#include <fcntl.h>
#include <unistd.h>
#include <stddef.h>
#include <getopt.h>

/* FreshTomato Standards */
#define TRX_MAGIC           0x30524448  /* "HDR0" */
#define NETGEAR_MAGIC       0x5E24232A  /* Netgear CHK */
#define TRX_MAX_LEN         0x4000000   /* 64MB - ARM Limit */

static uint32_t crc32_tab[256];
static int verbose = 0;

/* Mimics hndcrc32 logic in bcmutils/mtd */
static void init_crc32(void) {
	uint32_t i, j, c;
	for (i = 0; i < 256; i++) {
		for (c = i, j = 8; j > 0; j--)
			c = (c & 1) ? (0xEDB88320L ^ (c >> 1)) : (c >> 1);
		crc32_tab[i] = c;
	}
}

static uint32_t crc32_calc(uint32_t crc, const uint8_t *buf, size_t len) {
	while (len--) crc = crc32_tab[(crc ^ *buf++) & 0xFF] ^ (crc >> 8);
	return crc;
}

struct trx_header {
	uint32_t magic, len, crc32, flag_version, offsets[3];
};

static void usage(char *prog) {
	printf("Usage: %s [-v] -i <file1> [file2...]\n\n"
	       "Options:\n"
	       "  -i <file>   Firmware file(s) to check (TRX, CHK, BIN)\n"
	       "  -v          Verbose output (detailed header info)\n"
	       "  -h          Show this help\n\n"
	       "Returns: 0=Valid, 1=Format Error, 2=CRC Error, 3=Size Error, 4=I/O Error\n", prog);
}

static int perform_check(const char *fn) {
	int fd, res = 1;
	struct stat st;
	uint8_t *buf;
	uint32_t head_len;

	fd = open(fn, O_RDONLY);
	if (fd < 0) return 4;
	if (fstat(fd, &st) < 0 || st.st_size < sizeof(struct trx_header)) {
		close(fd);
		return 3;
	}

	/* mmap is optimal for routers: no extra RAM allocation needed for buffer */
	buf = mmap(NULL, st.st_size, PROT_READ, MAP_PRIVATE, fd, 0);
	if (buf == MAP_FAILED) {
		close(fd);
		return 4;
	}

	uint8_t *trx_ptr = NULL;
	uint32_t magic = *(uint32_t *)buf;

	/* Match mtd.c detection logic */
	if (magic == TRX_MAGIC) {
		trx_ptr = buf;
	} else if (magic == NETGEAR_MAGIC) {
		/* Netgear CHK: First 4-8 bytes contain header length (Big Endian) */
		head_len = __builtin_bswap32(*(uint32_t *)(buf + 4));
		if (head_len < st.st_size && *(uint32_t *)(buf + head_len) == TRX_MAGIC)
			trx_ptr = buf + head_len;
	} else {
		/* BIN/U2ND: Scan for TRX inside the wrapper (mimics mtd seek) */
		for (int i = 0; i < 512 && i < (st.st_size - 32); i++) {
			if (*(uint32_t *)(buf + i) == TRX_MAGIC) {
				trx_ptr = buf + i;
				break;
			}
		}
	}

	if (trx_ptr) {
		struct trx_header *h = (struct trx_header *)trx_ptr;
		if (h->len <= (st.st_size - (trx_ptr - buf)) && h->len <= TRX_MAX_LEN) {
			init_crc32();
			/* mtd-write2 logic: CRC starts at flag_version (offset 12) */
			uint32_t calc = crc32_calc(0xFFFFFFFF, (uint8_t *)&h->flag_version, h->len - 12);
			res = (calc == h->crc32) ? 0 : 2;

			if (verbose) {
				printf("\nFile: %s\n", fn);
				printf("  Format:   %s\n", (magic == TRX_MAGIC) ? "Native TRX" : (magic == NETGEAR_MAGIC) ? "Netgear CHK" : "BIN/U2ND Wrapper");
				printf("  TRX Size: %u bytes\n", h->len);
				printf("  Header CRC: 0x%08X | Calc: 0x%08X\n", h->crc32, calc);
				printf("  Status:   %s\n", (res == 0) ? "VALID" : "INVALID (CRC Mismatch)");
			}
		} else {
			res = 3; /* Too large or truncated */
		}
	}

	if (!verbose) {
		printf("%s: %s\n", fn, (res == 0) ? "VALID" : "INVALID");
	}

	munmap(buf, st.st_size);
	close(fd);
	return res;
}

int trx_check_main(int argc, char *argv[]) {
	int ch, i, final_res = 0, found_i = 0;
	verbose = 0;
	optind = 0; /* Reset for multicall binary */

	while ((ch = getopt(argc, argv, "vhi:")) != -1) {
		switch (ch) {
			case 'v': verbose = 1; break;
			case 'h': usage(argv[0]); return 0;
			case 'i': 
				found_i = 1;
				/* Process all files following -i */
				for (i = optind - 1; i < argc; i++) {
					if (strcmp(argv[i], "-i") == 0) continue;
					int r = perform_check(argv[i]);
					if (r != 0) final_res = r;
				}
				break;
		}
	}

	if (!found_i) {
		usage(argv[0]);
		return 1;
	}

	return final_res;
}
```

HommeOursPorc:
@wellloaded suggested to add a test function directly into mtd.c instead. I didn't want to break anything so I didn't went that way, but the AI think it would be a way shorter implementation like 30 lines which is quite tempting IMO. 

-------------------
Gemini: 
Integrating the check directly into mtd.c is actually a brilliant idea. In the world of embedded systems and router firmware, "less is more."
By adding a test flag to mtd, you eliminate the need for a separate applet, shared logic, and extra entries in the rc applet table. Since mtd already has the headers, the CRC functions (hndcrc32), and the I/O logic included, you only need about 20–30 lines of code to create a "test mode."
Here is how we can implement -t (test) inside mtd.c to achieve exactly what you want.
1. Update the Usage Message
Locate the usage() function or the help output in mtd.c and add:
fprintf(stderr, "       mtd-write2 -t file\n");
2. Implementation Logic in mtd.c
You can insert this logic into the main loop of mtd. The goal is to run the existing validation path but exit before the MEMERASE or write syscalls happen.
/* Add this logic into the part of mtd.c that handles mtd-write2 */

if (test_only) {
    printf("Testing firmware integrity: %s\n", path);
}

/* Existing mtd logic to find TRX starts here... */
if (magic == TRX_MAGIC || magic == NETGEAR_MAGIC) {
    // ... (Existing code that identifies the TRX offset)
    
    // Perform the CRC check exactly like mtd already does for writing
    uint32_t crc = 0xFFFFFFFF;
    // ... (Running the hndcrc32 calculation)

    if (crc != trx.crc32) {
        fprintf(stderr, "%s: Bad CRC (Calculated: 0x%08X, Expected: 0x%08X)\n", path, crc, trx.crc32);
        return 2; // Error code for Shell
    }

    if (test_only) {
        printf("%s: VALID\n", path);
        if (verbose) {
            printf("  Size: %u\n  Magic: 0x%08X\n", trx.len, trx.magic);
        }
        return 0; // Success - Exit before writing to flash!
    }
}

Why this is a "Pro Move":
 * Zero Redundancy: You stop carrying two versions of the CRC logic. If the CRC logic in mtd.c changes for a new hardware revision, your "test" tool is updated automatically.
 * Size Savings: You save the overhead of a second .o file and the string table entries in the rc binary. Every kilobyte counts in the /sbin/rc multi-call binary.
 * 100% Accuracy: You aren't "mimicking" the flash process anymore—you are using the flash process code. If mtd -t says it's valid, you are guaranteed that mtd write will accept it.
 * Shell Integration: You can still use it in scripts:
   if mtd-write2 -t /tmp/firmware.trx; then
    echo "Safe to flash!"
fi

Is it a good idea?
Yes. It is a much cleaner "Unix-style" approach. Instead of a new tool, you are extending a mature tool with a "dry-run" capability.
Would you like me to help you map out the specific line numbers in your mtd.c.txt where the getopt and the test_only flag should be inserted?
------------------------

So, mtd-wtite2 -t

OR

mtd-check2 ft.trx

I think having a different command to prevent accidental flash would be better. 

---------------------
Gemini: 

To keep the binary size as small as possible and avoid duplicating logic, the best approach is to add a new "entry point" or applet handle to mtd.c.
In FreshTomato, mtd-write2 is just a symbolic link to the rc multi-call binary, which then calls the code in mtd.c. By adding mtd-check2, we create a clean, dedicated command that shares 100% of the mtd code base without adding a separate file.
Here is the perfect, optimized implementation directly for your mtd.c.
1. The Logic Integration (Add to mtd.c)
This function uses the exact same hndcrc32 and header-skipping logic as the real flash process. It uses a small 8KB buffer, making it extremely "Low RAM" friendly.
```c
/* Add this function to mtd.c */
int mtd_check2_main(int argc, char *argv[])
{
	struct trx_header trx;
	uint32_t crc = 0xFFFFFFFF;
	uint8_t buf[8192];
	int c, v = 0;
	char *path = NULL;
	FILE *fp;
	size_t n;
	uint32_t magic;

	while ((c = getopt(argc, argv, "v")) != -1) {
		if (c == 'v') v = 1;
	}

	if (optind < argc) path = argv[optind];

	if (!path) {
		fprintf(stderr, "usage: mtd-check2 [-v] firmware.trx\n");
		return 1;
	}

	if ((fp = fopen(path, "r")) == NULL) {
		perror(path);
		return 4;
	}

	/* Find TRX Magic (Mimic mtd-write2 container skipping) */
	safe_fread(&magic, 1, 4, fp);
	fseek(fp, 0, SEEK_SET);

	if (magic == NETGEAR_MAGIC) {
		fseek(fp, 58, SEEK_SET); /* Skip CHK header */
	} else if (magic != TRX_MAGIC) {
		/* BIN/U2ND Search: Scan first 512 bytes for TRX */
		for (int i = 0; i < 512; i++) {
			if (fread(&magic, 1, 4, fp) != 4) break;
			if (magic == TRX_MAGIC) {
				fseek(fp, -4, SEEK_CUR);
				break;
			}
			fseek(fp, -3, SEEK_CUR);
		}
	}

	if (safe_fread(&trx, 1, sizeof(struct trx_header), fp) != sizeof(struct trx_header) || trx.magic != TRX_MAGIC) {
		fprintf(stderr, "%s: INVALID (No TRX magic found)\n", path);
		fclose(fp);
		return 1;
	}

	if (v) {
		printf("File: %s\nSize: %u bytes\nExpected CRC: 0x%08X\n", path, trx.len, trx.crc32);
	}

	/* CRC starts at flag_version (offset 12) per mtd.c spec */
	fseek(fp, ftell(fp) - sizeof(struct trx_header) + 12, SEEK_SET);
	size_t remaining = trx.len - 12;

	while (remaining > 0) {
		n = fread(buf, 1, (remaining > sizeof(buf)) ? sizeof(buf) : remaining, fp);
		if (n <= 0) break;
		crc = hndcrc32(buf, n, crc);
		remaining -= n;
	}
	fclose(fp);

	if (crc == trx.crc32) {
		if (v) printf("Calculated CRC: 0x%08X\nResult: VALID\n", crc);
		else printf("%s: VALID\n", path);
		return 0;
	} else {
		if (v) printf("Calculated CRC: 0x%08X\nResult: INVALID (CRC Mismatch)\n", crc);
		else printf("%s: INVALID\n", path);
		return 2;
	}
}
```

2. Linking it to the Shell
To make mtd-check2 work as a command, you need to register it in the router/rc/rc.c (or equivalent) applet table.
In your applet table (usually in rc.c):
{ "mtd-check2", mtd_check2_main },

Why this is the "Perfect" solution:
 * Size Efficiency: By putting this in mtd.c, you use the existing #include <trxhdr.h> and hndcrc32 function already linked in. It adds almost zero weight to the binary.
 * Safety: It mimics the exact file-pointer movement mtd-write2 uses to skip Netgear and Linksys headers. If mtd-check2 says it's valid, the router will definitely be able to flash it.
 * One-Line Output: By default, it outputs firmware.trx: VALID.
 * Verbose Mode: -v gives you the technical breakdown (Size, Expected vs. Calculated CRC).
 * Shell Error Codes:
   * 0: Valid
   * 1: Format Error (No TRX found)
   * 2: CRC Mismatch (Corrupt)
   * 4: I/O Error (File not found)
Usage:
 * mtd-check2 firmware.trx
 * mtd-check2 -v firmware.trx
 * echo $? (to see the result code)
My Opinion: The mtd-check2 handle is a much cleaner choice. It keeps the "writing" command and the "checking" command distinct for the user, preventing accidental flashes while keeping the code perfectly optimized under the hood.
Would you like me to show you exactly where to insert the applet entry in your rc.c table?
----------------

I can close this and work on that alternative solution. I'd like to have your opinions guys: @pedro0311 @MarsTomato @wellloaded

wellloaded:
You can close and re-open... or simply modify the commit linked to the PR.

What I suggest you do, instead of printing a human-readable output, is to have mtd-write2 spitting out an exit code. This would be good in general for the flashing part, but if -t can call the file integrity only, that also can be 0=OK 1=issue/s. Technically speaking you can print out more exit codes like 2,3,4,etc, each linked to a particular reason (each with the own error message).

which is pretty much what AI was telling here:
```
Shell Error Codes:

    0: Valid
    1: Format Error (No TRX found)
    2: CRC Mismatch (Corrupt)
    4: I/O Error (File not found)
    Usage:
```


Still, I would keep this as simple as possible.

HommeOursPorc:
> You can close and re-open... or simply modify the commit linked to the PR.
> 
> What I suggest you do, instead of printing a human-readable output, is to have mtd-write2 spitting out an exit code. This would be good in general for the flashing part, but if -t can call the file integrity only, that also can be 0=OK 1=issue/s. Technically speaking you can print out more exit codes like 2,3,4,etc, each linked to a particular reason (each with the own error message).
> 
I don't like this idea because mtd-write is called after killing all the services and if the firmware is bad you absolutely need to reboot the router at this point and for some connection type the router might be unreachable at this point too. It would be way better to catch the bad firmware before the "point of no return". This applies to both my script and the original tomato upgrade code.

> which is pretty much what AI was telling here:
> 
> ```
> Shell Error Codes:
> 
>     0: Valid
>     1: Format Error (No TRX found)
>     2: CRC Mismatch (Corrupt)
>     4: I/O Error (File not found)
>     Usage:
> ```
> 
> Still, I would keep this as simple as possible.

I would need to check the code. mtd write do output when the flash is successful, but in case of a bad flash I'm not sure. I think it doesn't output anything at all and as for the return code, I never checked. I'll check it out right now. 

Oh! It does output info and have a return code.
```bash
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# mtd-write2
usage: mtd-write2 [path] [device]
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# mtd-write2 test.trx linux
test.trx: File is too small (5 bytes)
root@AC68U-Relais:/tmp/home/root/usbr/firmwares# echo $?
255
```

HommeOursPorc:
I will close this and come back with an other proposition later.

HommeOursPorc:
Damn it I can't get make rc to work 😡😡😡

```
/root/freshtomato-arm/release/src-rt-6.x.4708/toolchains/hndtools-arm-linux-2.6.36-uclibc-4.5.3/bin/../lib/gcc/arm-brcm-linux-uclibcgnueabi/4.5.3/../../../../arm-brcm-linux-uclibcgnueabi/bin/ld: rc: hidden symbol `__aeabi_idiv' in /root/freshtomato-arm/release/src-rt-6.x.4708/toolchains/hndtools-arm-linux-2.6.36-uclibc-4.5.3/bin/../lib/gcc/arm-brcm-linux-uclibcgnueabi/4.5.3/libgcc.a(_divsi3.o) is referenced by DSO
/root/freshtomato-arm/release/src-rt-6.x.4708/toolchains/hndtools-arm-linux-2.6.36-uclibc-4.5.3/bin/../lib/gcc/arm-brcm-linux-uclibcgnueabi/4.5.3/../../../../arm-brcm-linux-uclibcgnueabi/bin/ld: final link failed: Bad value
collect2: ld returned 1 exit status
make: *** [Makefile:137: rc] Error 1

root@debian:~/freshtomato-arm/release/src-rt-6.x.4708/router/rc# make rc EXTRALDFLAGS="-lgcc_s"
 [rc] CC rc
/root/freshtomato-arm/release/src-rt-6.x.4708/toolchains/hndtools-arm-linux-2.6.36-uclibc-4.5.3/bin/../lib/gcc/arm-brcm-linux-uclibcgnueabi/4.5.3/../../../../arm-brcm-linux-uclibcgnueabi/bin/ld: rc: hidden symbol `__aeabi_idiv' in /root/freshtomato-arm/release/src-rt-6.x.4708/toolchains/hndtools-arm-linux-2.6.36-uclibc-4.5.3/bin/../lib/gcc/arm-brcm-linux-uclibcgnueabi/4.5.3/libgcc.a(_divsi3.o) is referenced by DSO
/root/freshtomato-arm/release/src-rt-6.x.4708/toolchains/hndtools-arm-linux-2.6.36-uclibc-4.5.3/bin/../lib/gcc/arm-brcm-linux-uclibcgnueabi/4.5.3/../../../../arm-brcm-linux-uclibcgnueabi/bin/ld: final link failed: Bad value
collect2: ld returned 1 exit status
make: *** [Makefile:137: rc] Error 1
```

Anyway 😒

PR #235, opened 2026-03-01
GUI: advanced-adblock.asp (v2) - Define blocklist in MB

wellloaded:
- Updated the adblock limit field to convert MB to Bytes
- Changed display units to MB
- Updated Notes section
- Updated script to use spin.svg


pedro0311:
Thanks!

PR #236, opened 2026-03-01
GUI:  Scan SSID/BSSID improvement

wellloaded:
- basic+virtual: New SSID Scan function
- basic+virtual: New BSSID scan function
- basic+virtual: Rearranged options and adjusted indent
- basic+virtual: Hiding for certain Wireless Modes e.g. where Auto is required this is now hidden
- basic+virtual: New Shared Key [Random] option allowing different password length
- basic+virtual: Introduced *some* special characters in the [Random] generated password
- basic+virtual: WDS - renamed “Automatic” to “Open” and “Link With...” to “Restricted”
- virtual: Hide tabs whose interface is not enabled under Overview
- virtual: Added SSID name to the tabs
- virtual: Added frequencies (e.g. / 2.4 GHz) to virtual interfaces and tab names
- virtual: Added for Wireless Client to get/display/set the WANx bridge mapping of reference
- virtual: Removed for Wireless Client any reference to LANx under bridge
- virtual: Renamed "Enable interface" to simply "Enable"
- virtual: Security parameters are now fully hidden if Security = Disabled (like the basic network does)

### DIFF
--- a/release/src-rt-6.x.4708/router/www/advanced-wlanvifs.asp
+++ b/release/src-rt-6.x.4708/router/www/advanced-wlanvifs.asp
@@ -33,7 +33,7 @@ var lastjiffiestotal = 0, lastjiffiesidle = 0, lastjiffiesusage = 100;
 
 <script>
 
-//	<% nvram("wl_auth,wl_auth_mode,wl_ap_isolate,wl_bss_enabled,wl_channel,wl_closed,wl_corerev,wl_crypto,wl_hwaddr,wl_ifname,wl_key,wl_key1,wl_key2,wl_key3,wl_key4,wl_lazywds,wl_mode,wl_nband,wl_nbw_cap,wl_nctrlsb,wl_net_mode,wl_passphrase,wl_phytype,wl_radio,wl_radius_ipaddr,wl_radius_key,wl_radius_port,wl_security_mode,wl_ssid,wl_vifs,wl_wds,wl_wds_enable,wl_wep_bit,wl_wpa_gtk_rekey,wl_wpa_psk,wl_bss_maxassoc,wl_wme,lan_ifname,lan_ifnames,t_features,wl_macmode,wl_maclist");%>
+//	<% nvram("wl_auth,wl_auth_mode,wl_ap_isolate,wl_bss_enabled,wl_channel,wl_clap_hwaddr,wl_closed,wl_corerev,wl_crypto,wl_hwaddr,wl_ifname,wl_key,wl_key1,wl_key2,wl_key3,wl_key4,wl_lazywds,wl_mode,wl_nband,wl_nbw_cap,wl_nctrlsb,wl_net_mode,wl_passphrase,wl_phytype,wl_radio,wl_radius_ipaddr,wl_radius_key,wl_radius_port,wl_security_mode,wl_ssid,wl_vifs,wl_wds,wl_wds_enable,wl_wep_bit,wl_wpa_gtk_rekey,wl_wpa_psk,wl_bss_maxassoc,wl_wme,lan_ifname,lan_ifnames,wan_ifname,wan2_ifname,wan3_ifname,wan4_ifname,t_features,wl_macmode,wl_maclist");%>
 
 var cprefix = 'advanced_wlanvifs';
 var vifs_possible = [];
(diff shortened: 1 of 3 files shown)


pedro0311:
1. Generating a "Shared Key" regardless of the length setting always returns a string with 63 characters.
2. "Clear" doesn't work.
3. Why isn't "Clear" available in basic-network?

pedro0311:
Oh, one more: please do not use eval() when calling nvram until it's really necessary.

wellloaded:
> 1. Generating a "Shared Key" regardless of the length setting always returns a string with 63 characters.
> 
>     2. "Clear" doesn't work.
> 
>     3. Why isn't "Clear" available in basic-network?

Let me check on this as soon as I have a couple of spare hours.

wellloaded:
@pedro0311 thanks for the testing/feedback. 
Please try the latest commit version (updated in this PR already) which addresses the following:

1. Key always 63 characters -> Fixed in wireless.js. Refactored random_psk(): the select value is read and stored before resetting the dropdown to index 0, so the chosen length is always respected. Additionally, removed disabled from the "Random" placeholder <option> in both ASP files, which could cause selectedIndex = 0 reset to fail in some browsers.

2. Clear doesn't work -> Fixed in wireless.js. The old code called verifyFields(null, 1) after clearing, which re-validated the now-empty field and showed an error. Now it calls ferror.clear(e) instead, which properly clears both the field and any error state.

3. Clear not in basic-network -> Added <option value="clear">Clear</option> to both PSK dropdowns in basic-network.asp.

4. eval() for nvram -> Replaced all 27 eval('nvram.xxx'+u+'_yyy') calls with nvram['xxx'+u+'_yyy'] bracket notation in basic-network.asp. Also fixed 1 eval('fom.f_wl'+u+'_wepidx') → fom['f_wl'+u+'_wepidx']. Only the legitimate eval(text) for the AJAX wlscan callback remains.

5. Messed commit -> Branch rebased onto latest arm-master (be1a812). Single clean commit with exactly 3 files: basic-network.asp, advanced-wlanvifs.asp, wireless.js. No usb.c or other stray changes.

wellloaded:
BTW this is not a "heavy" change as such, and I believe a very useful one to have. I would push it down all the way to MIPS R1

lancethepants:
@wellloaded 
I've recently had to merge some very large commits like this one and I have a couple thoughts.

Please limit the scope of your commits to one purpose. This advertises two different changes, Scanning and Password changes which are not related to each other. They shouldn't be in a single commit. Also many many changes not even related to either of these things.

Please for all that is holy limit your changes to that one **specific** purpose.  I'm looking at **basic-network.asp** and knowing this is going to be nightmare to merge downstream.  All over the code base 
```
/* REMOVE-BEGIN */
if (hasUsbSupport) {
/* REMOVE-END */
```
Why are these being left in for a pull request?   This was done as well in advanced-vlan.asp which Pedro had to clean up later. These combined with all the random formatting changes are a nightmare for me to merge, and they have no place to be there.  All these non-related changing muddy up the water when trying to use tools like git blame. You lose context when trying to see when and why a particular line of code was changed. I even like how the vlan page looks now, but I did not like how it got there.

pedro0311:
@lancethepants : please wait for my merge, I'll simplify it a bit.

pedro0311:
@wellloaded 
You use (in various places) `wlX_clap_hwaddr`, while in FreshTomato only the `wl_clap_hwaddr` variable is defined in nvram and only for the RTN+ branch.

https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/tomato.c#L767

So these non-existent variables need to be added to nvram first, set the default value to "" in defaults.c, and only then use them in scripts, otherwise they won't even be even saved to nvram.

All their uses in scripts and functions must be surrounded by "RTNPLUS-BEGIN/RTNPLUS-END".

Unless I don't understand something here?

pedro0311:
@wellloaded : please note that I have already spent 3 weeks working on this commit, including creating a universal version of advanced-wlanvifs.asp that will allow the same commit to be applied to all branches.
https://freshtomato.org/downloads/freshtomato-arm/2026/2026.1/TESTS/236.patch

wellloaded:
> Please for all that is holy limit your changes to that one **specific** purpose. I'm looking at **basic-network.asp** and knowing this is going to be nightmare to merge downstream. All over the code base

That page was on overhaul I guess that's why, but I get your point, and thanks for the feedback.

> 
> ```
> /* REMOVE-BEGIN */
> if (hasUsbSupport) {
> /* REMOVE-END */
> ```
> 
> Why are these being left in for a pull request? This was done as well in advanced-vlan.asp which Pedro had to clean up later. These combined with all the random formatting changes are a nightmare for me to merge, and they have no place to be there. All these non-related changing muddy up the water when trying to use tools like git blame. You lose context when trying to see when and why a particular line of code was changed. I even like how the vlan page looks now, but I did not like how it got there.

The VLAN page redesign was complex due to accumulated code modifications that broke the original logic. I'm referring to decades of external contributions. The marks (REMOVE-BEGIN/END) on overview, VLAN, and virtual wireless pages are intentional and not leftovers. They enable mount --bind for testing without full compilation. They auto-remove post-compilation or when conditions aren't met, as browsers can't parse raw markup.

wellloaded:
> @wellloaded You use (in various places) `wlX_clap_hwaddr`, while in FreshTomato only the `wl_clap_hwaddr` variable is defined in nvram and only for the RTN+ branch.
> 
> https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/tomato.c#L767
> 
> So these non-existent variables need to be added to nvram first, set the default value to "" in defaults.c, and only then use them in scripts, otherwise they won't even be even saved to nvram.
> 
> All their uses in scripts and functions must be surrounded by "RTNPLUS-BEGIN/RTNPLUS-END".
> 
> Unless I don't understand something here?

This PR, like the VLAN and Wireless Survery is also from few months ago. I'll have to work out what I did that, but in general I would not expect this commit to require new variables at all. The core logic of the page is to do the same things, hopefully better, with the same resources we have.

pedro0311:
> I'll have to work out what I did that, but in general I would not expect this commit to require new variables at all.

You use this: `W('<input type="hidden" id="_wl'+u+'_clap_hwaddr" name="wl'+u+'_clap_hwaddr">');`

It won't work because there is no `wlX_clap_hwaddr` variable defined in nvram, but only one: `wl_clap_hwaddr`

pedro0311:
P.S. Please start with my version of the patch, as it is adapted to the latest changes in advanced-wlanvifs.asp.
At the same time, I'm not sure where to add RTNPLUS-BEGIN/END, even the ones I've already added.

pedro0311:
Just uploaded updated version of 236.patch.

pedro0311:
OK, I need to split this PR/commit down into:
1. Password
2. SSID/BSSID scan

Unfortunately, the SSID/BSSID part doesn't work.

I wanted to set up 'wireless client' using the VLAN page on my RT-N12U - see the inconsistencies in network <-> vlans - also on VLAN I should not be able to select WPA2 (only) in this mode for this router, only as it is on network - WPA/WPA2.

basic
advanced

wellloaded:
> OK, I need to split this PR/commit down into:
> 
>     1. Password
> 
>     2. SSID/BSSID scan
> 
> 
> Unfortunately, the SSID/BSSID part doesn't work.
> 
> I wanted to set up 'wireless client' using the VLAN page on my RT-N12U - see the inconsistencies in network <-> vlans - also on VLAN I should not be able to select WPA2 (only) in this mode for this router, only as it is on network - WPA/WPA2.
> 
> basic advanced

It works here in local (on my RT-AC1900P) when you say it doesn't work what exactly? 
Is the Scan not producing anything or what are problem are you facing?
Also (most importantly) does it work on any of your other devices?
Overall what I spotted is that the BSSID is not displayed on your version, might this have to do with the recent changes you applied to the page?

pedro0311:
Password part added: https://github.com/FreshTomato-Project/freshtomato-arm/commit/361e451436287ca51f9a0921c67d8eea3863edd8
Also updated (part2) with SSID/BSSID scan: https://freshtomato.org/downloads/freshtomato-arm/2026/2026.1/TESTS/236.patch

> Overall what I spotted is that the BSSID is not displayed on your version, might this have to do with the recent changes you applied to the page?

I can't test anything with the original patch, because all I have is js errors in console on RT-N12U because of lack of RTNPLUS-BEGIN/END stuff.
You can download the updated patch (part 2), apply it to the latest repo, and build an image for a MIPS R1/R2 router (like my RT-N12U).

pedro0311:
Updated patch once more.

wellloaded:
@pedro0311 where can I find the latest version of this? I can have a look to finish it off after your latest modifications.

pedro0311:
The last one is still there: https://freshtomato.org/downloads/freshtomato-arm/2026/2026.1/TESTS/236.patch

PR #244, opened 2026-03-13
robocfg.c - Implement link state persistence (bug workaround)

wellloaded:
Due to a bug in the robocfg.c its code was modified to introduce a hold timer of 5 sec when the port state changes state from connected (e.g. 1000FD) to DOWN. This is because tests demonstrated that a fake DOWN lasts between 2 and 4 seconds. It is possible this bug is HW sourced e.g. HW take too long to answer.


pedro0311:
https://github.com/FreshTomato-Project/freshtomato-arm/commit/f06226609e10699c104f2d42ef97bfac2ec5e7e9

PR #245, opened 2026-03-20
rc: wireguard.c: Insert firewall rules instead of appending. Also add a prerouting rule in case DMZ is being used.

lancethepants:


pedro0311:
Thanks!

PR #248, opened 2026-04-02
porthealth - fix for nvram variable and defaults

wellloaded:
It appears like the handling of the porthealth_cfg nvram variable was not part of the code that went into prod.

This fix is an attempt to:
- properly define the variable so that it can be read/written but .asp and script
- tweak the defaults on both nvram and .asp

In general the Max errors was too high (lowered from 50/min to 10/min) and the TTL should have been much higher (increased from 60 to 900 so 15 min).

### DIFF
--- a/release/src-rt-6.x.4708/router/httpd/tomato.c
+++ b/release/src-rt-6.x.4708/router/httpd/tomato.c
@@ -698,6 +698,9 @@ static const nvset_t nvset_list[] = {
 #ifdef TCONFIG_BCMNAT
 	{ "bcmnat_disable",		V_01				},
 #endif
+#ifdef TCONFIG_BCMARM
+	{ "porthealth_cfg",		V_LENGTH(0, 128)		},
+#endif
 /* advanced-vlan */
 	{ "vlan0ports",			V_TEXT(0, 17)			},
 	{ "vlan1ports",			V_TEXT(0, 17)			},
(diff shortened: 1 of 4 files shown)


pedro0311:
https://github.com/FreshTomato-Project/freshtomato-arm/commit/6d083e52ecbff5d33ec37bc7bf006fcf3903abde

lancethepants:
@pedro0311 
Thank you for simplifying it down to the actual meat and potatoes of the change.  Large sweeping code layout and formatting changes really complicate things for me.

wellloaded:
> @pedro0311 Thank you for simplifying it down to the actual meat and potatoes of the change. Large sweeping code layout and formatting changes really complicate things for me.

Indeed, two things though:

- the last commit I made, included what it looks to be a considerable amount of changes for the .asp only. This was introduced by my local studio code as I'm using a plugin called autopep8. This is automatically indenting based on parenthesis and other criteria. It really depends on how you see this, but it's either needing indentation attention or rather let's keep it as close as the original file as possible. I personally don't have an opinion on this but since this was made by a (I'm reading) respected plug-in I felt like leaving it. I'm also struggling with the EOF character for .asp only as @pedro0311 knows very well by now. AI suggested this might have to do with my github settings rather than the editor itself, but I'm yet to find a viable solution.

- About this particular feature, as T64 doesn't use Broadcom's robocfg you'll have to tweak the script to operate **ip link** or equivalent to enable/disable ports. The rest I'd expect to work out of the box.

pedro0311:
> Thank you for simplifying it down to the actual meat and potatoes of the change. Large sweeping code layout and formatting changes really complicate things for me.

I only ever use https://github.com/johnwait/notepad2-mod-jrb/ with the right settings and I encourage everyone to use it - without unnecessary bells and whistles and having everything under control.

PR #254, opened 2026-04-20
tomato.js - Add case 'display' to handle span rendering

wellloaded:
In the case where a table is composed by a mixture of input fields and read-only data, this new case 'display' helps handling the read-only information properly. It displays text in a cell without input fields, and doesn't prompt any input field when adding new records.

This allows for tables to be used in a more creative way like the MB column in this example:

			this.init('adblock-grid', '', 50, [
					{ type: 'checkbox', prefix: '<div class="centered">', suffix: '<\/div>' },
					{ type: 'text', maxlen: 130 },
					{ type: 'display', value: '', attrib: 'style="display:block;margin:0 auto"' },
					{ type: 'text', maxlen: 40 }
				]);

VrTy7ZlaHo

### DIFF
--- a/release/src-rt-6.x.4708/router/www/tomato.js
+++ b/release/src-rt-6.x.4708/router/www/tomato.js
@@ -1580,6 +1580,14 @@ TomatoGrid.prototype = {
 						else
 							s += '>';
 					break;
+						case 'display':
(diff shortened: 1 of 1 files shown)


pedro0311:
Thanks!

PR #257, opened 2026-04-26
Update advanced-misc.asp port health cache 900 default

scoobasteve83:
Either the default value is wrong being 9000 or the label for the default is wrong saying 900 one or the other.


pedro0311:
Thanks.

PR #258, opened 2026-04-28
Update advanced-misc.asp 1000mb full missing.

scoobasteve83:
Its been brought to the attention that 1000mb full is missing as an option while I get that autonegotiate should give you this speed I know what it is like to want to be able to select 1000mb as an option there is something about the implicit nature of selecting such a value. half isn't even really a thing for 1GB in production env.

And it is ruffling a few feathers not having the option.

I am unsure of where this is all handled downstream up stream or if it was something to save space.

But perhaps whe can call it 1000mb/auto in that case to make people feel a bit better if we are trying to do so it will give people the feeling of selecting 1000mb for option 4

### DIFF
--- a/release/src-rt-6.x.4708/router/www/advanced-misc.asp
+++ b/release/src-rt-6.x.4708/router/www/advanced-misc.asp
@@ -166,7 +166,7 @@ function save() {
 		for (i = 3; i <= 20; ++i) a.push([i, i + ' seconds']);
 		createFieldTable('', [
 			{ title: 'Boot Wait Time *', name: 'wait_time', type: 'select', options: a, value: fixInt(nvram.wait_time, 3, 20, 3) },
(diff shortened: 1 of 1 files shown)


pedro0311:
This isn't a good idea and doesn't make any sense.
If you set 'auto' on a router with 1 Gbps ports and you don't get that speed on a given port (checking it with status-ovierview, for example), you know something's wrong and you need to investigate the cause. If you set it to 1000HD, the status will show that value, but it will be fraught with connection issues (drops or other timeouts) – and instead of looking for problems with the cable/connection/device on the other end of the cable, you'll be left wondering pointlessly what's wrong.
Of course, the latest release of FT includes "Port Health," but that doesn't change the overall thinking.

scoobasteve83:
yes

PR #259, opened 2026-05-07
www: nas-ups.asp: Add async refresh functionality for UPS data

lancethepants:


pedro0311:
I've never tested UPS because I simply don't have it, and I rarely looked into this asp file, so I'm surprised that such crap was done there :P
Thanks!

PR #260, opened 2026-05-08
mdu: fix SIGSEGV and other fixes

lancethepants:
- curl_dump_cb: drop the trailing *out = '\0' (wrote one byte past libcurl's buffer when input had no CRs) and guard the trim checks with size > 0; the OOB write was corrupting libcurl's state and causing transfers to abort after the first response headers.
- _http_req: default *body = "" on entry so callers don't dereference uninitialized stack when curl returns an error code other than CURLE_OK / CURLE_RECV_ERROR with a 200 response code.
- update_cloudflare: null-check strchr before dereferencing it, so a malformed response missing the closing quote no longer segfaults.
- main: use a literal "%s" format with snprintf instead of passing an NVRAM value as the format string (CWE-134 / UB on '%' chars).

### DIFF
--- a/release/src-rt-6.x.4708/router/mdu/mdu.c
+++ b/release/src-rt-6.x.4708/router/mdu/mdu.c
@@ -385,14 +385,13 @@ static int curl_dump_cb(CURL *handle, curl_infotype type, char *data, size_t siz
 			++out;
 		}
 	}
-	*out = '\0'; /* null-terminate the modified string (for convenience) */
 
 	/* adjust the size after removing '\r' */
-	size = out - data; 
+	size = out - data;
 
-	if (data[size - 1] == '\n')
+	if (size > 0 && data[size - 1] == '\n')
 		size -= 1;
-	if (is_info && data[size - 1] == ':')
+	if (size > 0 && is_info && data[size - 1] == ':')
 		size -= 1;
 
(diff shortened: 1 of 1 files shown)


lancethepants:
musl libc is a bit more picky about things.  Also tested on r7000 and didn't seem to break anything.

pedro0311:
> Makefile: drop redundant -lssl -lcrypto and --start-group brackets.

I can't see it?

`			                inet_pton(AF_INET6, addr, &(struct in6_addr){0}) == 1))
#else
			                0))`

IIRC it will throw warnirngs/errors on MIPS gcc.

lancethepants:
I removed the comment on the makefile changes, that was my changes only.  I will probably remove that other part when I have sec, it wasn't a critical section of code to make things work for me.

lancethepants:
lol, now that I've sat down to really look at this, that errant change must have been me being out of sync with FT since 2026.2.  I just copied my version to my FT repo not realizing I was already so far behind your changes. I'll re-sync to your changes and resubmit.

pedro0311:
OK, so please let me know when it's ready.

pedro0311:
Added as a separate commit: https://github.com/FreshTomato-Project/freshtomato-arm/commit/f9cd6b49ea4b35933ad0271d3594416b9a682f20

PR #265, opened 2026-05-12
Update advanced-misc.asp

scoobasteve83:
add note to clarify autonegotiate.


pedro0311:
I would rather add such information to the wiki here: https://wiki.freshtomato.org/doku.php/advanced-misc

PR #266, opened 2026-05-14
httpd: do_file: allow paths with symlinks, some calling functions require this

lancethepants:


lancethepants:
This is the reason for the change.
https://www.linksysinfo.org/index.php?threads/tomato64-development-discussion-thread.78306/page-27#post-367104

pedro0311:
Right - I've never used it, so I completely forgot to check it...
Thanks!

PR #270, opened 2026-05-22
GUI: Ethernet port description field

wellloaded:
Y6mwRBsxXA

- enables an inline editing of Ethernet port description
- available on the status-overview.asp and advanced-vlan.asp
- max 8 chars per port description are allowed. Make it last :-)
- uses a single new variable "eth_desc"
- "%" is used internally, so it's not a valid character to be used for the description
- there is no save button, the text is written in nvram as soon as the focus moves elsewhere.
- it intentionally doesn't introduce a style modification, and it has been tested on basic and advanced themes

### DIFF
--- a/release/src-rt-6.x.4708/router/httpd/tomato.c
+++ b/release/src-rt-6.x.4708/router/httpd/tomato.c
@@ -573,6 +573,7 @@ static const nvset_t nvset_list[] = {
 
 	{ "lan_state",			V_01				},
 	{ "lan_desc",			V_01				},
+	{ "eth_desc", 			V_LENGTH(0, 80)			},
 	{ "lan_invert",			V_01				},
 	{ "lan_dhcp",			V_01				},	/* DHCP client [0|1] - obtain a LAN (br0) IP via DHCP */
 	{ "lan_proto",			V_WORD				},	/* static, dhcp */
@@ -1902,7 +1903,7 @@ static void _execute_command(char *url, char *command, char *query, wofilter_t w
 
 	/*
 	 * execute script via shell
-	 * NOTE: do NOT change to execvp(argv) – UI depends on full shell semantics
+	 * NOTE: do NOT change to execvp(argv) ï¿½ UI depends on full shell semantics
 	 */
 	snprintf(cmd, sizeof(cmd), "%s 2>&1", webExecFile);
 	web_pipecmd(cmd, wof);
--- a/release/src-rt-6.x.4708/router/shared/defaults.c
+++ b/release/src-rt-6.x.4708/router/shared/defaults.c
@@ -539,6 +539,7 @@ struct nvram_tuple router_defaults[] = {
 
 	{ "lan_state",			"1"				, 0 },	/* Show Ethernet LAN ports state (0|1) */
 	{ "lan_desc",			"1"				, 0 },	/* Show Ethernet LAN ports state (0|1) */
+	{ "eth_desc", 			""				, 0 },	/* Ethernet port custom descriptions (MAX_PORT_ID+1 entries, separated by %) */
 	{ "lan_invert",			"0"				, 0 },	/* Invert Ethernet LAN ports state (0|1) */
 
 	{ "mwan_num",			"1"				, 0 },
--- a/release/src-rt-6.x.4708/router/www/advanced-vlan.asp
+++ b/release/src-rt-6.x.4708/router/www/advanced-vlan.asp
@@ -23,7 +23,7 @@
 
 <script>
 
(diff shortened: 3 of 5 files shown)


pedro0311:
You forgot to add ",eth_desc" to nvram in status-overview.asp

pedro0311:
I meant in status-data.jsx

wellloaded:
Hum, but that shouldn't be needed right?
I mean, this info is straight from the nvram and not updating at every refresh.

pedro0311:
So where do you get the nvram "eth_desc" value on status-overview.asp?
I can't see it.

wellloaded:
Oh I see what you mean, in this particular case it's fetched like this:

`var ed = (nvram.eth_desc || '').split('%');`

Which, as I understand, it doesn't require the traditional nvram variable name to be declared at the top of the page.

pedro0311:
On each .asp page, you must first retrieve the value for a given variable before using it. You don't do that here, which is why I mentioned it.

wellloaded:
As I understand this part of the code:

Reading any nvram variable is always possible e.g. a declaration is not needed. For writing the nvram though it depends on how the page is structured.

The main difference between status-overview and advanced-vlan is that the former doesn't have a Save button nor verifyFields(). I think that's the key difference and also why eth_desc is defined under advanced-vlan but not under status-overview.

By all means, please try the commit as it is, it works on my side at least. If you prefer me to declare the variable I can do that, but without Save button I think there is nothing using it really?

scoobasteve83:
Would be cool if you could add a vlan description column so you could lable what the vlan number represents...

For instance if it is IOT you can see it visually when you segment your network enough I think you'll come to the same conclusion I do networking daily and I still forget from time to time what number is which with only 6VLAN's  would be more useful than a single port label imho but :/ who knows you can visually look and see what is plugged into what with a port with the VLAN label its a bit different its like ok which LAN is that again what number oh crap better check DHCP reservation so I untag to the right network.... I've put laptops on the wrong IOT network or camera network by mistake I can't count how many times and couldn't figure out why it wasn't connecting lol its cause those networks are locked down and don't allow DHCP if it is not reserved. But I get them confused with my test net on and off. Which number is which. If I don't use my test vlan for a bit I forget which number coincides and guess the wrong one. 

Not the end of the world but being that you are on labels this is a very common thing that switches and other decent equipment normally allows labeling of VLAN even a lot of the cheap stuff allows it.

HommeOursPorc:
> Would be cool if you could add a vlan description column so you could lable what the vlan number represents...

I like that idea. I vote for this 👍

pedro0311:
> Which, as I understand, it doesn't require the traditional nvram variable name to be declared at the top of the page.

Wrong. nvram.eth_desc does nothing here (it's seen by javascript as 'undefined').
You have to use `<% nv("eth_desc"); %>` or add it to the <% nvram(""); %> declaration in status-data.jsx.
Example (t_model_name) here: https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/www/wwan-sms.asp#L158

wellloaded:
It's now added, but I still don't understand why it worked without it :-)

scoobasteve83:
Thanks for helping educate @pedro0311 wisdom is nothing if it isn't shared.  @rs232 has been giving a good push lately fun to see what he comes up with.

pedro0311:
Heh, not that way either ;)
To know what I'm talking about, you can open for example: ROUTER_IP/status-data.jsx?_http_id=TID_your_tid_here (check TID in console -> F12).
As you can see, variable nvram is defined there as a array of:
```
nvram = {
'variable1': 'value',
'variable2': 'value',
'variable3': 'value',
/.../
};
```

If you enter `<% nvram("single_variable"); %>` in status-overview.asp, you will replace the already existing (large) nvram array with a single entry:
```
nvram = {
'single_variable': 'value'
}
```
and nothing else! All other needed nvram variables on status-overview will be "undefined" to JS!
So as I said already: add ',eth_desc' to status-data.jsx in already defined nvram there, or as a single variable `<% nv(""); %>` but I would prefer the first one.

scoobasteve83:
Because when you create a large variable string then you have to chop and separate it and it turns into a headache cause everything you grab needs that routine even though it is a way to save nvram space if needed as each variable label takes up space in nvram as well 1byte per char on avg is this correct?

Thats why you prefer the first method right?

pedro0311:
Let me explain a bit: `<%` and `%>` are special characters that are interpreted by the asp parser (https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/parser.c#L29) as, "this is the beginning of a block of code that will do something/return some value on the page, not a regular html/JS".
In the case of `nvram()` it will be an array of variables from NVRAM, defined as I wrote above.
`nvram()` is nothing else than `asp_nvram()` here: https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/nvram.c#L164 whose appropriate call is defined here: https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/httpd/tomato.c#L131 .

So the latter is an array: name_called_on_asp -> name_of_function_in_c.

`nv()`, on the other hand, returns only and exclusively a "pure", single variable from NVRAM in the form of a string.

wellloaded:
@pedro0311 whenever you have a min, please check the updated commit.

status-data.jsx ->

`//	<% nvram("eth_desc,router_name,wan_domain,wan_hostname,wan_dns,lan_hwaddr,lan_proto,lan_ipaddr,dhcpd_startip,dhcpd_endip,lan_netmask,wl_security_mode,wl_crypto,wl_mode,wl_wds_enable,wl_hwaddr,wl_net_mode,wl_radio,wl_channel,lan_gateway,wl_ssid,wl_closed,wl_nband,t_model_name,t_features,lan_ifname,lan_ifnames,wan_ifnames,tomatoanon_enable,tomatoanon_answer,lan_desc,wan_ppp_get_ip,wan_pptp_dhcp,wan_pptp_server_ip,wan_ipaddr_buf,wan_gateway,wan_gateway_get,wan_get_domain,wan_hwaddr,wan_ipaddr,wan_netmask,wan_proto,wan_run_mtu,wan_sta,mwan_num,wan_modem_type,wan_hilink_ip,wan_status_script,mwan_cktime,wan_weight,wan_ck_pause,dnscrypt_proxy,dnscrypt_priority,stubby_proxy,stubby_priority"); %>`

pedro0311:
OK - FYI, next time if you want to add a new variable name to 'nvram(...)', put it at the end. ;)

pedro0311:
Thanks!

PR #271, opened 2026-05-23
GUI: status-overview.asp - WAN icons fix

wellloaded:
This fixes a little bug in the status-overview.asp Ethernet port naming when no active Ethernet ports are present.
Essentially with no Ethernet it assumes all the ports are LAN only, this modification reinforces the logic checking the VLAN configuration instead and displaying the WANs as expected.

Before:
uDeR7J8DSV

After:
796Okg1f3H

### DIFF
--- a/release/src-rt-6.x.4708/router/www/status-overview.asp
+++ b/release/src-rt-6.x.4708/router/www/status-overview.asp
@@ -221,6 +221,22 @@ function c(id, htm) {
 	E(id).cells[1].innerHTML = htm;
 }
 
+function calcWanPortCount() {
+	var count = 0;
+
+	for (var uidx = 1; uidx <= nvram.mwan_num; ++uidx) {
+		var u = (uidx > 1) ? uidx : '';
+		if ((nvram['wan'+u+'_proto'] != 'disabled')
+/* USB-BEGIN */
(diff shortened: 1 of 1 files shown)


pedro0311:
Thanks!

krystian-r:
I've got the same question about it: on NETGEAR R6400 v2 ARM model there are LAN form 0 to 3 ports and WAN0 port. Is it possible to rename WAN0 to WAN1 and LAN ports start counting from 1 to 4?

wellloaded:
> I've got the same question about it: on NETGEAR R6400 v2 ARM model there are LAN form 0 to 3 ports and WAN0 port. Is it possible to rename WAN0 to WAN1 and LAN ports start counting from 1 to 4?

I think this is a big change. Logical port names, for whatever reason, were designed to start from 0 and the whole firmware is built around it.

I think this goes as far as storing 0 in some variables etc. So I'd expect such a change to be very intrusive in the existing code.

krystian-r:
> I think this is a big change. Logical port names, for whatever reason, were designed to start from 0 and the whole firmware is built around it. I think this goes as far as storing 0 in some variables etc. So I'd expect such a change to be very intrusive in the existing code.

OK. I thought so, but I wasn't sure. I had to ask for knowing is it possible or not. I understand there are many dependencies related to these descriptions. Thanks for reply!

PR #274, opened 2026-05-28
rc: dnsmasq.c: fix dual-MAC DHCP reservations dropped by strict ether_atoe()

lancethepants:


pedro0311:
It is an old one, thanks!

PR #275, opened 2026-05-30
rc: firewall.c: fix multi-second boot stall from start_firewall holding its own lock

lancethepants:
the run_*_firewall_script() scripts handle their own locking


pedro0311:
Thank you!

PR #277, opened 2026-06-11
www: admin-iptraffic.asp: fix missing firewall restart on first cstats enable

lancethepants:


pedro0311:
@lancethepants : don't we need a similar fix for bwm/rstats?

lancethepants:
Only cstats requires special firewall rules (-m account) to gather its information.  rstats just reads its info from the /proc/net/dev kernel interface to gather its information.

pedro0311:
So let me ask this differently: in admin-bwm.asp (here: https://github.com/FreshTomato-Project/freshtomato-arm/blob/arm-master/release/src-rt-6.x.4708/router/www/admin-bwm.asp#L203) we restart rstats and the firewall in normal mode, but a few lines below, for "new" mode, we restart only rstats with the "new" suffix.
So from what you wrote, restarting the firewall in normal mode is redundant and we can remove it?

lancethepants:
I can't see that there's any reference to "firewall" on the admin-bwm.asp page. It looks like in both instances that only rstats is restarted.

pedro0311:
Right, thanks!

pedro0311:
(well, I think I should see an ophthalmologist...)

PR #282, opened 2026-07-02
dnsmasq: check for infinite (non-expiring) leases. Explicitly cast to unsigned long for 64bit time_t on 32bit systems

lancethepants:


pedro0311:
Thanks!

PR #291, opened 2026-07-26
www: status-log.asp: Enhance search with multiple terms and regex support

HommeOursPorc:
WARNING this is 100% AI generated. Very sorry about that, but JS is like Chinese to me 😛

Refactor search functionality to support multiple terms and regex highlighting.

This commit make the log search function smarter and can now search multiple terms and also accept condition "|" (or) and -string (search minus this term).

Example:

DHCP PC
will find every line that contains both terms

DHCP|PC
Will find every line that contains DHCP or PC

DHCP -PC
Will find every line that contains DHCP but not PC

it works well with words

Search term: support
Will find every word including "support" including "supported"

search: support -ed
Will find every word "support" but will exclude "supported"

I'm sorry for the AI mess. I don't know if this is worth your time Pedro, but I'm sharing because it works well. 

You can test the code with mount --bind

### DIFF
--- a/release/src-rt-6.x.4708/router/www/status-log.asp
+++ b/release/src-rt-6.x.4708/router/www/status-log.asp
@@ -23,7 +23,8 @@
 var cprefix = 'status_log';
 
 var currentSearch = '';
-var negativeSearch = 0;
+var currentSearchTerms = [];
+var currentHighlightRegex = null;
 var currentFilterValue = 0;
 var currentlyScrolling = false;
 var scrollingDetectorTimeout;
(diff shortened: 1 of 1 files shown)


HommeOursPorc:
I've played with this some more and found a little issue, it won't search the log level, but it will highlight it. 

If I search "DHCP", I get 840 results
If I search "info", I get 0 results
If I search "DHCP INFO", it won't return any result,
If I search "DHCP|INFO", it will return all the result for DHCP (840 results) and highlight the info log level. 

No big deal IMO. 

Screenshot_20260726_234454_Firefox


Little detail about the "-" syntax,

Searching for "support" will give all support|supports|supported results. 

Screenshot_20260726_234650_Firefox

Searching for "support -ed" will give incorrect results. 

Screenshot_20260726_234705_Firefox

Searching for "support -supported" is correct though.

Screenshot_20260726_234827_Firefox

Anyway, its your call Pedro if this is a good idea or not. Sorry again for the AI mess. I'd like to code like you... Maybe in an other life lol.

pedro0311:
Thank you for the contribution and for proposing the extended log search functionality. ;)

Based on this PR, I prepared a revised implementation here: https://github.com/FreshTomato-Project/freshtomato-arm/commit/9c21d60c6a684f7f84628256df5de2b047483d88 .

The new commit keeps the same general idea, but reduces the amount of code and explicitly supports:

* space-separated terms as `AND`
* `|` as `OR` within a term group
* `-` as exclusion
* searching across all displayed columns, including the log level
* highlighting positive literal matches
* safe rendering through DOM text nodes instead of `innerHTML`

I decided not to merge this PR directly because the original implementation had a few issues:

* the log level was highlighted but was not included in the actual row filtering;
* the description suggested regular-expression support, while the implementation performed literal matching;
* exclusion applied to the entire row, which made examples such as `support -ed` behave differently from what the description implied;
* quoted phrase parsing added complexity but was not documented and could produce matches spanning separate table columns;
* the overall implementation was larger than necessary for the intended feature set.

Your PR was still useful as the basis for the final implementation. Thanks!

HommeOursPorc:
WOW! You're on fire lately Pedro! Very nice! 

I didn't look at the code, I would've not understand it anyway. It was completely AI generated and I anticipated that it would be... Bad...

I used Gemini pro, it made it on the first try (one question, one answer)! I tested it and it worked "correctly" without breaking anything, so I decided to share the code. I wasn't sure if it was worth it. I'm happy you could do something out of it 😃

Thank you for your hard work!!!

Issue #149, opened 2025-10-10
Web Admin: allow using the same port for internal and external access


bobyang3:
I have Local Access configured to use HTTPS on port 4443 in Web Admin. I’d like to enable Remote Access so I can help my parents manage the router without using a VPN, which tends to be slower. However, the web interface doesn’t allow me to set Remote Access to use port 4443 as well. It would be great if this was allowed, as I can configure it via SSH to use the same port for both. Additionally, since SSH allows using the same port for internal and external access, Web Access should support this as well on the web ui. Thank you!

<img width="625" height="134" alt="Image" src="https://github.com/user-attachments/assets/33fc6289-c595-4cc6-83f6-0a9c5d78ee97" />

HommeOursPorc:
You can forward the port with SSH if you allow external access with SSH.

From eibgrad on the linksysinfo forum:
https://www.linksysinfo.org/index.php?threads/plz-hlp-no-web-interface.78834/post-353406

```
ssh -L <local-port>:<remote-ip>:<remote-port> <user>@<ddns-domain-name>

Example:
ssh -L 4443:192.168.1.1:4443 <user>@<my-ddns-domain>
```

Now you can access the router's GUI using http://localhost:4443 without having to expose the GUI directly.

HommeOursPorc:
This is probably **NOT** secure, but this is what I do with my test router inside my own LAN. 

Add this to Administration > Scripts > Firewall
`iptables -I INPUT -p tcp --dport 80 -j ACCEPT`

And forward the port too.

<img width="2827" height="1241" alt="Image" src="https://github.com/user-attachments/assets/8d1b7759-4e78-4e3c-ab36-31889eb361b7" />

Like this I can access the GUI on port 80 from the WAN side, in my case it is still inside my LAN. But again, this is probably a security risk. Use at your own risk 😉

bobyang3:
Thank you @HommeOursPorc ! Actually, I didn’t need to do all those steps. I just had to SSH in and run the following commands — after that, everything worked perfectly: (on the web, I set 4443 for local, and 4444 as remote first to get around the error pops)

```
nvram set https_wanport=4443
nvram commit
```

However, there’s an issue: whenever I make any changes on that page, eg, "Open Menus", the same error pops up again. I then have to repeat the same SSH steps to reset https_wanport.

bermudahonk:
[...]

PR #297, opened 2026-08-05
www: admin-access.asp: Fix using same port for remote and local management and also remove standard port (80/443) restriction (issue #149)

HommeOursPorc:
Fix for issue #149

This commit remove the port 80/443 restriction and impose a TWO stage confirmation that the user must acknowledge to accept the risk.

It also remove the same port restriction for remote and local management and add a warning message that the user need to acknowledge to continue with this configuration.

Also updated the little port note accordingly.

Let me know if this is OK @pedro0311


pedro0311:
https://github.com/FreshTomato-Project/freshtomato-arm/commit/de4da81bbd2c876adf28905d361cfa63e4e98db4

HommeOursPorc:
Hmmm this was more complicated than I initially thought. Thanks for fixing my mess @pedro0311 😁

PR #305, opened 2026-08-18
rc: qos.c: create the QOSSIZE chain whenever a size rule needs it

lancethepants:
The QOSSIZE chain was created under "if (!sizegroup)", but sizegroup is also incremented for a layer7 rule, where it only means "every rule from here on must stay reclassifiable" and no chain is involved.

A layer7 rule followed by a rule with a bounded transfer range therefore emitted "-A QOSSIZE" rules for a chain that was never declared, and since iptables-restore is all or nothing, the whole mangle table failed to load and QoS did not come up at all.

Track whether the chain has been created in its own flag.


pedro0311:
Thanks!

PR #308, opened 2026-08-25
Fix Docker builds

mobrembski:
-Updated Docker image to Debian 13, along with its packages -
-Disable Samba parallel build - its broken
-Update my mail :)

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM debian:12
+FROM debian:13
 
 ENV DEBIAN_FRONTEND=noninteractive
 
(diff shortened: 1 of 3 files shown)


pedro0311:
> -Disable Samba parallel build - its broken

Are you sure about this? I have no problem compiling samba on Debian 13 with parallel build enabled.

mobrembski:
> > -Disable Samba parallel build - its broken
> 
> Are you sure about this? I have no problem compiling samba on Debian 13 with parallel build enabled.

Try to do it from fresh build. It will fail, and another shot (without cleaning) will work. Annoying. 
At least what i see on Docker.

pedro0311:
OK, thanks!

PR #311, opened 2026-09-04
others: mwwatchdog: gate on active WAN count, fix mwanJob toggle and PID file release

lancethepants:
Use the leading run of non-disabled WANs (mirrors mwan_active_num()) instead of the configured mwan_num, so routes, the wan1 prefix rewrite and mwanroute recovery only engage when MultiWAN is really up; mwanJob now adds/removes the alive job by state instead of flip-flopping it on every run, and the PID file is only removed by the run that owns it (add/del/alive never take the lock).


pedro0311:
I forgot about this one, thanks!

PR #312, opened 2026-09-10
httpd/rc/www: re-encode mojibake comment characters as UTF-8

lancethepants:
non-UTF-8 characters can mess up some search tools


pedro0311:
Thanks!

